PR #6, opened 2024-01-15
Add CI check for correctly formatted Go code

danielg0:


danielg0:
I've added a check so that code that would be changed by running `go fmt` fails the CI pipeline. There's also `go vet` as well, which runs a bunch of static analyzers on the code to check for common errors - might be worth incorporating too?

aDotInTheVoid:
Yeah, go vet is worthwhile but not needed for this

r=me when your happy with this

PR #8, opened 2024-01-16
Add Go coverage checking

danielg0:
Coverage checking requires generating a coverage profile during the test step, then we can filter out the total coverage from it.

Go also has the ability to generate a HTML coverage report. By uploading the coverage profile as an artifact, a new workflow (run when CI fails), can pull down that artifact, generate a coverage report and upload that.


aDotInTheVoid:
What does this look like on failure?

aDotInTheVoid:
https://github.com/gpuctl/gpuctl/actions/runs/7541270482

danielg0:
Hmm, I've fixed the yaml file itself, and I didn't see a syntax error that time, but it still didn't run. Will it not til it's merged in?

aDotInTheVoid:
Dont see why that should be. Might be better off just printing to console for now.

danielg0:
> What does this look like on failure?



This isn't a failure case, but when there is failure, you can go in and see coverage at a per-function and global coverage percentages

PR #10, opened 2024-01-16
Implement Frontend Skeleton & Fetching of Stats

### LINKED ISSUE #16: Retrieve all machine stats from API and display on frontend
Technical Spec:
- [x] `api/stats/all`
- [x] Decide on some standard JSON format
- [x] Refresh stats after period of time
- [x] Go API Server (The Go Programming Language/linked tutorials on Discord)
- [x] TSX webpage shit that displays the data
- [x] React shit that fetches the data from the backend
- [x] Tests

NathanielB123:
Closes https://github.com/gpuctl/gpuctl/issues/16

Technically there are no tests for the API back-end, but this is a simplest-possible implementation just to check the front-end works (i.e: it just returns static JSON). I'm not sure what unit tests you could write for it honestly.


frontend/src/Utils/Hooks.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { EffectCallback, useEffect, useState } from "react";
+import { discard } from "./Utils";
+
+/**
+ * Daniel named this
+ */
+export const useJarJar = <T,>(
+  f: () => Promise<T | null>
+): [T | null, EffectCallback] => {
+  const [v, setV] = useState<T | null>(null);
```
Euphoride: assuming this hook is used for async state management when sending api calls

would suggest that rather than the 
```
  T | null
```

type, we use a more structured approach, i.e.:

```
type UseJarJarState<T> = {
  data: T | null;
  isLoading: boolean;
  error: Error | null;
};
```

and also, it might be worth building in utils an explicit optional type with functorial/monadic properties if the T | null pattern is expected to come up often

NathanielB123: I've done a little cooking in https://github.com/gpuctl/gpuctl/pull/10/commits/0cd32348191381685fe5739f65b7c2227786364e

(Haven't gone so far as creating an `Optional`/`Maybe` yet as IMO `| null` is more ergonomic in TypeScript (no patter matching, and I think use-cases for nested `Maybe`s are somewhat limited - we usually just want to flatmap anyway, which `| null` just kinda gives you for free)

NathanielB123: Though... on the pattern matching side of things...

```ts
enum MTag {
  Just = "Just",
  Nothing = "Nothing",
}

type Just<T> = {
  tag: MTag.Just;
  data: T;
};

type Nothing = {
  tag: MTag.Nothing;
};

type Maybe<T> = Just<T> | Nothing;

const just = <T,>(x: T): Just<T> => ({ tag: MTag.Just, data: x });
const nothing = (): Nothing => ({ tag: MTag.Nothing });

// Motive/Eliminator terminology inspired by theorem proving/dependent types
// Arguably this is not *quite* the same thing, so I'm happy to take other suggestions
type MaybeMotive<T, U> = {
  just: (x: T) => U;
  nothing: () => U;
};

const maybeElim = <T, U>(m: Maybe<T>, motive: MaybeMotive<T, U>): U => {
  switch (m.tag) {
    case MTag.Just: {
      return motive.just(m.data);
    }
    case MTag.Nothing: {
      return motive.nothing();
    }
  }
};

const example = () => {
  const foo: Maybe<boolean> = just(true);

  maybeElim(foo, {
    just: (x) => x,
    nothing: () => false,
  });
};
```
having a record with a field for each case looks remarkably similar to a switch statement lol

NathanielB123:
```ts
maybeElim(foo, {
  just: (x) => x,
  nothing: () => false,
});
```
vs (fantasy-land TypeScript with pattern matching)
```ts
case foo of {
    just(x) => x,
    nothing() => false,
}
```
it's ***almost*** there (though no nested patterns) - so maybe if we wanna do FP-style stuff, we could design our APIs for ADT-like things like this?

Issue #15, opened 2024-01-16
Write infra for getting GPU info from nvidia-smi


aDotInTheVoid:
**User Story**

I want to view GPU stats (i.e. usage) conveniently in order to see if it is in use.

**Technical Spec**

- [x] Call into NVidia SMI
- [x] Parse XML
- [x] Store in internal data structure
- [x] Print (and test that it prints correctly)

PR #22, opened 2024-01-17
Satellite XML parsing and filtering

YRMYJASKA:
Capabilities for pulling information from `nvidia-smi` and parsing thereof. Currently satellite binary will just print out the parsed status every 5 seconds to the terminal. No support for processes nor multiple GPUs (as no way to test that at the moment)

NB: naming and general conventions are not set in stone, did not think that long about them. 

Related issue: #15


aDotInTheVoid:
Looks like you need to run `go fmt`.

Ideally your ide does this automaticly, so you dont need to make a new commit when ci fails, but thats fine here.

YRMYJASKA:
> Looks like you need to run `go fmt`.
> 
> Ideally your ide does this automaticly, so you dont need to make a new commit when ci fails, but thats fine here.

yeah, agreed. It does run semi-automatically but emacs is having a proper issue with zealously replacing some spaces with tabs for ONE SPECIFIC FILE. Fixing atm.

internal/status/handlers/nvidia_test.go
```diff
@@ -0,0 +1,69 @@
...
+            sp := strings.Split(filename, ".")
+            resloc := testDataRoot + "/" + sp[0] + ".json"
+            dump, err := ioutil.ReadFile(resloc)
+            if err != nil {
+                t.Fatalf("Could not read test result data at %s: %v", resloc, err)
+            }
+            expected = dump
+        }
+
+        j = append(j, 10) // HACK: annoying newline at the end of stored data...
```
aDotInTheVoid: NIT: As go constants/literals are untyped, you can do

```go
var x byte = '\n'
```

so we avoid a magic number here.

internal/status/handlers/nvidia.go
```diff
@@ -0,0 +1,351 @@
...
+		return 0, nil
+	}
+	return strconv.ParseInt(parts[0], 10, 0)
+}
+
+// Helper function to unmarshal Nvidia XML dump
+func ParseNvidiaSmi(input []byte) (NvidiaSmiLog, error) {
+	var result NvidiaSmiLog
+	if e := xml.Unmarshal(input, &result); e != nil {
+		return result, e
```
aDotInTheVoid: NIT: This should be `NvidiaSmiLog{}, e`, so we're sure the data is zero

internal/status/handlers/nvidia.go
```diff
@@ -0,0 +1,351 @@
...
+	if isDirty {
+		return res, errors.New("Parsing of data fields failed.")
+	}
+
+	return res, nil
+}
+
+// Helper function for extracting GPU data such as available memory and fan speed
+func parseUIntWithUnit(input string) (uint64, error) {
+	parts := strings.Split(input, " ")
```
aDotInTheVoid: It looks like we're just discarding the unit here, which feels bad.

If we genuinly don't need it, it's nicer to use [strings.Cut](https://pkg.go.dev/strings#Cut), although that might be nicer anyway

internal/status/handlers/nvidia.go
```diff
@@ -0,0 +1,351 @@
...
+		MemoryTotal:       memoryTotal,
+		MemoryUtilisation: memUtil,
+		GPUUtilisation:    gpuUtil,
+		MemoryUsed:        memoryUsed,
+		FanSpeed:          fanSpeed,
+		Temp:              temp}
+
+	// report back catastrophic failures
+	if isDirty {
+		return res, errors.New("Parsing of data fields failed.")
```
aDotInTheVoid: This error should be in a global var, so users can use `errors.Is` and friends.

```go
// global
var ErrBadField error = errors.New("...")

// in a function
   return res, ErrBadField
```

internal/status/handlers/nvidia.go
```diff
@@ -0,0 +1,351 @@
...
+		UsedMemory        string `xml:"used_memory"`
+	} `xml:"process_info"`
+}
+
+// Filter down the relevant information from our nvidia-smi dump
+func (xml NvidiaSmiLog) FilterStatus() (status.GPUStatusPacket, error) {
+	fanSpeed, err := parseUIntWithUnit(xml.Gpu.FanSpeed)
+	isDirty := err != nil
+	memoryTotal, err := parseUIntWithUnit(xml.Gpu.FbMemoryUsage.Total)
+	isDirty = isDirty || err != nil
```
aDotInTheVoid, line 277: NIT: This can probably use `||=`

internal/status/handlers/nvidia_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+        j, err := json.Marshal(r)
+        if err != nil {
+            t.Errorf("Could not marshal status packet to JSON: %v (file %s)", err, fileloc)
+            continue
+        }
+        // Read expected json
+        var expected []byte
+
+        {
```
aDotInTheVoid: NIT: Why's this needed?

internal/status/handlers/nvidia_test.go
```diff
@@ -0,0 +1,69 @@
...
+            resloc := testDataRoot + "/" + sp[0] + ".json"
+            dump, err := ioutil.ReadFile(resloc)
+            if err != nil {
+                t.Fatalf("Could not read test result data at %s: %v", resloc, err)
+            }
+            expected = dump
+        }
+
+        j = append(j, 10) // HACK: annoying newline at the end of stored data...
+        if !reflect.DeepEqual(expected, j) {
```
aDotInTheVoid: NIT: This can just be [bytes.Equal](https://pkg.go.dev/bytes#Equal), to avoid needing to pull in reflection

internal/status/handlers/_testdata/nvidia/ash03.xml.faulty
```diff
@@ -0,0 +1,2 @@
+Failed to initialize NVML: Driver/library version mismatch
```
aDotInTheVoid: What's the point of having this checked in?

internal/status/gpustatus.go
```diff
@@ -0,0 +1,16 @@
+package status
+
```
aDotInTheVoid, line 2: Eventually this should move into it's own package, but we don't need to do that now.

internal/status/gpustatus.go
```diff
@@ -0,0 +1,16 @@
...
+type GPUStatusPacket struct {
+	// Contextual information
+	Name          string `json:"gpu_name"`
+	Brand         string `json:"gpu_brand"`
+	DriverVersion string `json:"driver_ver"`
+	MemoryTotal   uint64 `json:"memory_total"`
+
+	// Temporal statistics
+	MemoryUtilisation uint64 `json:"memory_util"` // Percentage of memory used
```
aDotInTheVoid, line 11: What's the point in renaming all these fields for JSON?

YRMYJASKA: just so that we use snake_case for JSON. This was purely a stylistic choice on my part

internal/status/handlers/_testdata/nvidia/ash03.xml.faulty
YRMYJASKA: because it can happen and we want to catch that somehow. I just added tests that do check for those failures and that they fail accordingly.

internal/status/handlers/nvidia_test.go
YRMYJASKA: expected output is stored in a JSON file in `_testdata`, so we need to read that?

internal/status/handlers/nvidia.go
YRMYJASKA: ~this does not exist in Go~ will look into it. go was complaining before

YRMYJASKA:
@aDotInTheVoid Happy to merge?

danielg0:
Closes #15

PR #25, opened 2024-01-17
Added interface as requested

YRMYJASKA:


YRMYJASKA:
@danielg0 can you review?

danielg0:
Looks good - bit more context to this: I was unsure what the split was between the expected data source interface and the public nvidia-specific functions.

As a note for potential future refactors, I'd be in favour of moving this whole module "handlers" to be a submodule of satellite (as it's only used here)

PR #26, opened 2024-01-17
Define basic database abstraction layer

### LINKED ISSUE #13: Set up link to postgres db
Module interfacing between groundstation and database.

- [x] Interface defining functions implemented by the abstraction layer
- [x] Postgres implementation using the built-in SQL library
- [x] Mock layer that we can use for tests

Going to assume all error-checking is done, we're given some set of structs/enums and pass them straight on.

https://pkg.go.dev/database/sql

Database interface needs:
- Heartbeat: given a struct of performance metrics from a satellite, update that row/push a new datapoint into the db
  - What if we don't recognise the machine?
- User: given a user id, get their usage across all machines
- All: get the latest metrics for all "approved" machines
    - when we first receive a message from a machine, do we want to add it immediately, or have some function for adding/ignoring it?
    - Probably fine if we only install the satellite binary through the groundstation web interface, but in theory it can always be installed out of band
- Any more suggestions, do add them...

I think we should be fine without a specific "Get data for one machine" endpoint. We're dealing with only a couple machines, and data that is largely numeric/text-based. If anyone disagrees, do say

danielg0:
Define and implement the set of functions that the postgres database provider implements

I'm doing this generally using interfaces to make it easy to write a mocking database interface for testing other components

Closes #13

### DIFF
--- a/cmd/groundstation/database/interface.go
+++ b/cmd/groundstation/database/interface.go
@@ -4,9 +4,6 @@ import "github.com/gpuctl/gpuctl/internal/status"
 
 // define set of operations on the database that any provider will implement
 type Database interface {
-	// initialise the database
-	Initialise() error
-
 	// add the info from a heartbeat packet to the data store
 	// that machine may have not been seen before
 	Heartbeat(packet status.GPUStatusPacket) error
--- a/cmd/groundstation/database/interface.go
+++ b/cmd/groundstation/database/interface.go
@@ -5,7 +5,7 @@ import "github.com/gpuctl/gpuctl/internal/status"
 // define set of operations on the database that any provider will implement
 type Database interface {
 	// update the last seen time for a satellite to the current time
-	UpdateLastSeen(host String) error
+	UpdateLastSeen(host string) error
 
 	// record a new data point for a satellite in the data store
 	AppendDataPoint(packet status.GPUStatusPacket) error
--- a/cmd/groundstation/database/interface.go
+++ b/cmd/groundstation/database/interface.go
@@ -4,10 +4,12 @@ import "github.com/gpuctl/gpuctl/internal/status"
 
 // define set of operations on the database that any provider will implement
 type Database interface {
-	// add the info from a heartbeat packet to the data store
-	// that machine may have not been seen before
-	Heartbeat(packet status.GPUStatusPacket) error
+	// update the last seen time for a satellite to the current time
+	UpdateLastSeen(host String) error
+
+	// record a new data point for a satellite in the data store
+	AppendDataPoint(packet status.GPUStatusPacket) error
 
 	// get the latest metrics for all approved machines
-	AllMachines() ([]status.GPUStatusPacket, error)
+	LatestData() ([]status.GPUStatusPacket, error)
 }
--- a/cmd/groundstation/database/interface.go
+++ b/cmd/groundstation/database/interface.go
@@ -1,0 +1,16 @@
+package database
+
+import "github.com/gpuctl/gpuctl/internal/status"
+
+// define set of operations on the database that any provider will implement
+type Database interface {
+	// initialise the database
+	Initialise() error
+
+	// add the info from a heartbeat packet to the data store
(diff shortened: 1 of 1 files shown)


danielg0:
I've defined the basic interface I'm going to write functions to, so I'm going to move on to implementation. If anyone does have strong opinions on database schema, I think now would be the best time to express them

cmd/groundstation/database/interface.go
```diff
@@ -0,0 +1,13 @@
+package database
+
+import "github.com/gpuctl/gpuctl/internal/status"
+
+// define set of operations on the database that any provider will implement
+type Database interface {
+	// add the info from a heartbeat packet to the data store
+	// that machine may have not been seen before
+	Heartbeat(packet status.GPUStatusPacket) error
```
aDotInTheVoid: Wat????!!???

The heartbeat just contains "This node is up", and no GPU data. It's not a concept at the DB layer.

Also, the GPU data as stored in the Database shouldn't be related to the GPU data as tranmitted over the network.

Euphoride: about the first point, i agree

about the second point, gpu data in the database is a function of the gpu data sent by our satellites, and actually we may just have a table which is verbatim the gpu stats sent to us. once schema is established we'll have more info on this but our database is likely going to be tables where either:

table.extend(f(gpu stat received))

or 

table.extend(reducer([list of gpu stats received]))

danielg0: I agree with the first point in the long term, though I think it could be useful whilst we get the system up and running to have one endpoint that pushes gpu status info into the database, then eventually split that out into heartbeats that just update us that the server is still alive and other channels for more secure data later.

As for the second point, I had reused that struct because it's the both the form the data is generated/reaches us from the satellites (https://github.com/gpuctl/gpuctl/blob/2404956106d3388dfbd1ddccacbeac00c83863fd/internal/status/gpustatus.go#L3) and the form of data that the frontend expects (https://github.com/gpuctl/gpuctl/blob/1f4d700707db9a2ad8026a19261a05a4eda1e0a8/frontend/src/App.tsx#L7), so it seemed sensible that we would want to store it in that form too

aDotInTheVoid (approved):

cmd/groundstation/database/interface.go
```diff
@@ -0,0 +1,15 @@
...
+// define set of operations on the database that any provider will implement
+type Database interface {
+	// update the last seen time for a satellite to the current time
+	UpdateLastSeen(host string) error
+
+	// record a new data point for a satellite in the data store
+	AppendDataPoint(packet status.GPUStatusPacket) error
+
+	// get the latest metrics for all approved machines
+	LatestData() ([]status.GPUStatusPacket, error)
```
Euphoride, line 14: Probably needs our schema to be defined a little bit clearer, but the latest data might be variable depending on what is requested by the front-end in terms of view on the database. It would be nice to have a way to execute santisied SQL queries against the database and return data according to the schema's data model.

danielg0: In what sort of ways do you think we might want to vary the query? I can see it being easy to add arguments to `LatestData()` so we can get the latest data for a particular machine or set of machines (shared or workstation for instance - specified by constants) - would this be enough?

aDotInTheVoid:
can we just merge this now?

we’ve been spinning on this, and i feel like it’d be more useful to change it later when writing the application code that uses it, rather than try to speculate what would be useful

danielg0:
Oh, maybe I've misunderstood - my understanding was that this pr would contain the postgres implementation (which I'm currently working on), as well as tests of its behaviour

aDotInTheVoid:
Oh amazing. A postgres layer would be very useful at this point. Didn't realize that was in the works (i think this is my fault for messing up standup).

danielg0:
Having spoken with Mila irl yesterday and seen your pr approval, I'm happy with the actual interface definition if you want to merge this in now and then I'll do the postgres implementation as another pr

PR #27, opened 2024-01-18
Initial REST api for heartbeat.

### LINKED ISSUE #14: HTTP Heartbeat from Satellite to Groundstation
Assume cleartext and that the groundstation address/url is constant/hardcoded.

- [ ] Sat: Initiate http connection with groundstation
- [ ] Transmit with basic information
- [ ] On recieve, just log for now
    - One day this will go into the database

aDotInTheVoid:
Closes #14

This need a bit more work before being usefull, but I want to land/get your review now, because groundstation<->satellite comunication is important.

Core Ideas:

- `uplink` package defines data format for this communication
  - Any types that need to cross the network here should be defined here.
  - IE we should move `GPUStatsPacket` into here.
- `fempto` contains abstractions for HTTP.

To test:

```go
go run ./cmd/groundstation/
# in another tab:
go run ./cmd/satellite/
```

Which will output something like:

```go
2024/01/18 01:51:05 INFO Starting groundstation
2024/01/18 01:51:08 INFO New Request req_no=3 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60420
2024/01/18 01:51:08 INFO New Request req_no=1 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60404
2024/01/18 01:51:08 INFO New Request req_no=2 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60374
2024/01/18 01:51:08 INFO Received a heartbeat req_no=2 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=5 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60394
2024/01/18 01:51:08 INFO Received a heartbeat req_no=1 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=4 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60436
2024/01/18 01:51:08 INFO Received a heartbeat req_no=3 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO Received a heartbeat req_no=4 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO Received a heartbeat req_no=5 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=6 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60380
2024/01/18 01:51:08 INFO Received a heartbeat req_no=6 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=7 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60432
2024/01/18 01:51:08 INFO Received a heartbeat req_no=7 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=8 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60454
2024/01/18 01:51:08 INFO Received a heartbeat req_no=8 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=9 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60396
2024/01/18 01:51:08 INFO Received a heartbeat req_no=9 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
2024/01/18 01:51:08 INFO New Request req_no=10 method=POST url=/gs-api/heartbeat/ from=127.0.0.1:60442
2024/01/18 01:51:08 INFO Received a heartbeat req_no=10 prev_time=0001-01-01T00:00:00.000Z cur_time=2024-01-18T01:51:08.362Z
```

(This probably needs to be automated)

### DIFF
--- a/cmd/groundstation/main.go
+++ b/cmd/groundstation/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log/slog"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/gpuctl/gpuctl/cmd/groundstation/config"
@@ -27,23 +28,44 @@ func main() {
 	http.HandleFunc("/api/remote", remote.HandleStatusSubmission)
 	http.ListenAndServe(config.PortToAddress(configuration.Server.Port), nil)
 
-	mux := new(femto.Femto)
-	gs := groundstation{
-		lastSeen: make(map[string]time.Time),
-	}
-	femto.OnPost(mux, uplink.HeartbeatUrl, gs.heartbeat)
+	srv := NewServer()
+
+	http.ListenAndServe(":8080", srv)
 
 	// TODO: Make this configurable
-	err = mux.ListenAndServe(":8080")
+	// err = mux.ListenAndServe(":8080")
 	slog.Info("Shut down groundstation", "err", err)
+}
+
+type Server struct {
+	mux *femto.Femto
+	gs  groundstation
+}
+
+func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	s.mux.ServeHTTP(w, r)
+}
+
+func NewServer() *Server {
+	mux := new(femto.Femto)
+	gs := groundstation{lastSeen: make(map[string]time.Time)}
+
+	/// Register routes.
+	femto.OnPost(mux, uplink.HeartbeatUrl, gs.heartbeat)
+
+	return &Server{mux, gs}
 }
 
 type groundstation struct {
 	lastSeen map[string]time.Time
+	mu       sync.Mutex
 }
 
 func (gs *groundstation) heartbeat(data uplink.HeartbeatReq, req *http.Request, log *slog.Logger) error {
 	// TODO: Pull out just the IP here.
+	gs.mu.Lock()
+	defer gs.mu.Unlock()
+
 	from := req.RemoteAddr
 	prev := gs.lastSeen[from]
 
--- a/cmd/groundstation/main.go
+++ b/cmd/groundstation/main.go
@@ -2,24 +2,54 @@ package main
 
 import (
 	"fmt"
+	"log/slog"
 	"net/http"
+	"time"
 
 	"github.com/gpuctl/gpuctl/cmd/groundstation/config"
 	"github.com/gpuctl/gpuctl/cmd/groundstation/remote"
+	"github.com/gpuctl/gpuctl/internal/femto"
+	"github.com/gpuctl/gpuctl/internal/uplink"
 )
 
 func main() {
-	fmt.Print("Hey, world :3")
+	slog.Info("Starting groundstation")
 
 	configuration, err := config.GetConfiguration("config.toml")
 
 	if err != nil {
 		// TODO: Using logging library for auditing, fail soft
 		fmt.Println("Error detected when determining configuration")
+		return
 	}
 
+	// TODO: Move this into fempto.
 	http.HandleFunc("/api/remote", remote.HandleStatusSubmission)
-
 	http.ListenAndServe(config.PortToAddress(configuration.Server.Port), nil)
 
+	mux := new(femto.Femto)
+	gs := groundstation{
+		lastSeen: make(map[string]time.Time),
+	}
+	femto.OnPost(mux, uplink.HeartbeatUrl, gs.heartbeat)
+
+	// TODO: Make this configurable
+	err = mux.ListenAndServe(":8080")
+	slog.Info("Shut down groundstation", "err", err)
+}
+
+type groundstation struct {
+	lastSeen map[string]time.Time
+}
+
+func (gs *groundstation) heartbeat(data uplink.HeartbeatReq, req *http.Request, log *slog.Logger) error {
+	// TODO: Pull out just the IP here.
+	from := req.RemoteAddr
+	prev := gs.lastSeen[from]
+
+	gs.lastSeen[from] = data.Time
+
+	log.Info("Received a heartbeat", "prev_time", prev, "cur_time", data.Time)
+
+	return nil
 }
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -1,28 +1,43 @@
 package main
 
 import (
-	"encoding/json"
-	"fmt"
-	"github.com/gpuctl/gpuctl/internal/status/handlers"
+	"log/slog"
 	"time"
+
+	"github.com/gpuctl/gpuctl/internal/femto"
+	"github.com/gpuctl/gpuctl/internal/uplink"
 )
 
 func main() {
-	// Just get the GPU data and print it in JSON
-	gpuHandler := handlers.NvidiaGPUHandler{}
-	for {
-		res, err := gpuHandler.GetGPUStatus()
-		if err != nil {
-			fmt.Printf("Failed to parse: %v\n", err)
-			return
-		}
 
-		ser, err := json.Marshal(res)
-		if err != nil {
-			fmt.Printf("Failed to parse: %v\n", err)
-			return
-		}
-		fmt.Printf("%s\n", ser)
-		time.Sleep(5 * time.Second)
+	log := slog.Default()
+
+	s := satellite{
+		// TODO: Make this configurable
+		gsAddr: "http://localhost:8080",
 	}
+
+	log.Info("Starting satellite")
+
+	for i := 0; i < 10; i++ {
+		err := s.sendHeartBeat()
+		if err != nil {
+			log.Error("failed to send heartbeat", "err", err)
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	log.Info("Stopped satellite")
+
+}
+
+type satellite struct {
+	gsAddr string
+}
+
+func (s *satellite) sendHeartBeat() error {
+	return femto.Post(
+		s.gsAddr+uplink.HeartbeatUrl,
+		uplink.HeartbeatReq{Time: time.Now()},
+	)
 }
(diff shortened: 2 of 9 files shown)


cmd/groundstation/main.go
```diff
@@ -1,11 +1,41 @@
...
+}
+
+type groundstation struct {
+	lastSeen map[string]time.Time
+}
+
+func (gs *groundstation) heartbeat(data uplink.HeartbeatReq, req *http.Request, log *slog.Logger) error {
+	// TODO: Pull out just the IP here.
+	from := req.RemoteAddr
+	prev := gs.lastSeen[from]
```
aDotInTheVoid, line 70: This is racy. But no point fixing until it can be reproduced in a test. Easy enough to reproduce.

```go
toSpawn := 10000
	wg.Add(toSpawn)
	for i := 0; i < toSpawn; i++ {
		go func() {
			err := s.sendHeartBeat()
			if err != nil {
				log.Error("failed to send heartbeat", "err", err)
			}
			wg.Done()
		}()
	}
	wg.Wait()
```

but needs infra to spin up a test server

danielg0: Hey Alona, I agree that this is currently racy, but would I be right in thinking it would be solved once we're putting last seen into the database rather than storing it in a data structure shared by all the request handlers?

cmd/satellite/main.go
```diff
@@ -1,28 +1,43 @@
...
+}
+
+type satellite struct {
+	gsAddr string
+}
+
+func (s *satellite) sendHeartBeat() error {
+	return femto.Post(
+		s.gsAddr+uplink.HeartbeatUrl,
+		uplink.HeartbeatReq{Time: time.Now()},
```
danielg0, line 41: Rather than send the satellite's local time at the point it sent the heartbeat, might we want to instead store in the database the time at which we recieve the heartbeat at the groundstation?

This way, we're only using/storing timestamps from one source, and don't have to handle the clocks of satellites being/becoming out of sync with each other and with the groundstation's clock

danielg0:
Other than comments made above, this all looks good to me

cmd/groundstation/main.go
aDotInTheVoid: Yeah, the DB layer should fix this specific instance.

However I still want to land tests for this, as it'll force me to write enough of a harness to write a failing test (with the race detector).

cmd/satellite/main.go
aDotInTheVoid: That's probably right

internal/femto/femto.go
```diff
@@ -0,0 +1,68 @@
...
+
+	log.Info("New Request", "method", r.Method, "url", r.URL, "from", r.RemoteAddr)
+
+	if r.Method != http.MethodPost {
+		log.Info("Wanted GET, returned 405")
+		http.Error(w, "Expected GET", http.StatusMethodNotAllowed)
+		return
+	}
+
+	// TODO: Ensure all field's are present. I'm so mad this is hard.
```
Euphoride, line 39: I think checking that all fields are present should be okay to do - check my PR on #28

internal/femto/femto.go
```diff
@@ -0,0 +1,68 @@
...
+func doPost[T any](f *Femto, w http.ResponseWriter, r *http.Request, handle HandlerFunc[T]) {
+
+	reqNo := f.nextReqNo()
+	log := f.logger().With(slog.Uint64("req_no", reqNo))
+
+	log.Info("New Request", "method", r.Method, "url", r.URL, "from", r.RemoteAddr)
+
+	if r.Method != http.MethodPost {
+		log.Info("Wanted GET, returned 405")
+		http.Error(w, "Expected GET", http.StatusMethodNotAllowed)
```
Euphoride: Might be my 3am brain, but are we expecting gets or posts here?

internal/uplink/heartbeat.go
```diff
@@ -0,0 +1,11 @@
+// package uplink defines the types used in groundstation <-> satiate transport.
+
+package uplink
+
+import "time"
+
+const HeartbeatUrl = "/gs-api/heartbeat/"
+
+type HeartbeatReq struct {
+	Time time.Time
```
Euphoride, line 10: I think @danielg0 might have mentioned this already, but using the time that a request hits the actual server's endpoint is likely ideal here. Otherwise, time has no ground truth.

cmd/satellite/main.go
```diff
@@ -1,28 +1,43 @@
...
-			fmt.Printf("Failed to parse: %v\n", err)
-			return
-		}
 
-		ser, err := json.Marshal(res)
+	log := slog.Default()
+
+	s := satellite{
+		// TODO: Make this configurable
+		gsAddr: "http://localhost:8080",
```
Euphoride, line 17: I'd suggest that this reads from a configuration file for now, until we have DNS-based service discovery that's automated. ~~Maybe we need to build in a way to parameterise our satellites through a TOML file?~~ See #28.

cmd/satellite/main.go
```diff
@@ -1,28 +1,43 @@
...
+	}
+
+	log.Info("Starting satellite")
+
+	for i := 0; i < 10; i++ {
```
Euphoride, line 22: We probably want this to run as a coroutine in the background while the satellite does other work.

cmd/groundstation/main.go
Euphoride: Do we want to fix this by making the `lastSeen` map thread-safe, or by building in a thread-safe event queue (or using Go channels)?

cmd/groundstation/main.go
```diff
@@ -1,11 +1,41 @@
...
+	slog.Info("Starting groundstation")
+
+	mux := new(femto.Femto)
+	gs := groundstation{
+		lastSeen: make(map[string]time.Time),
+	}
+	femto.OnPost(mux, uplink.HeartbeatUrl, gs.heartbeat)
+
+	// TODO: Make this configurable
+	err := mux.ListenAndServe(":8080")
```
Euphoride: We should probably handle this error and log a sad path for audit purposes

Euphoride (commented):
Brief review on the PR as it is now

cmd/groundstation/main.go
aDotInTheVoid: Current approach: Slap a big mutex over it. Doing event queues is much more fragile.

(ask me about the horrors of the conversation event pump sometime :))

internal/femto/femto.go
aDotInTheVoid: that linked PR doesn't do this:

```go
func TestBuildStatusObjectEmpty(t *testing.T) {
	t.Parallel()

	emptyObj := []byte("{}")
	obj, err := buildStatusObject(emptyObj)
	assert.NoError(t, err)
	assert.Zero(t, obj)
}
```

passes (and it shouldn't)

internal/femto/femto.go
aDotInTheVoid: No, this is my 3am brain

cmd/satellite/main.go
aDotInTheVoid: At some point, but this can be done in the future.

internal/femto/femto.go
Euphoride: Huh, annoying. I saw this edge case on toml deserialisation too, although I fixed that. Seems like this is an edge case that can be manually guarded against in buiildStatusObject if this is the only thing that needs to be checked for.

With that in mind, why on earth does deserialisation fail on all other field mismatch cases but not the empty case? These semantics are weird...

aDotInTheVoid: it also passes on the missing fields case:

```go
func TestBuildStatusObjectMissingFields(t *testing.T) {
	t.Parallel()

	emptyObj := []byte(`{"gpu_name": "foo"}`)
	obj, err := buildStatusObject(emptyObj)
	assert.NoError(t, err)
	assert.Equal(t, "foo", obj.Name)
}
```

passes

internal/uplink/heartbeat.go
aDotInTheVoid: punted to #33, because this is more about comunication than accurate timekeeping

cmd/satellite/main.go
aDotInTheVoid: I'd rather do this with a command line arg at the start, but yeah.

internal/femto/femto.go
Euphoride: That seems unusual as #28 has a test that verifies this doesn't pass - TestBuildMalformedObject. At the moment this fails as expected. I'll check this other case tomorrow, but why does it pass if the other one fails as expected?

aDotInTheVoid: `TestBuildMalformedObject` passes because we have ``DriverVersion string `json:"driver_ver"``, but in JSON it's a number: `"driver_ver": 1.0`

so the error is `json: cannot unmarshal number into Go struct field GPUStatusPacket.driver_ver of type string`

the unmarhall failed, but not due to missing fields. i guess we should assert on the errors contents, not just that it's non-nil

cmd/satellite/main.go
Euphoride: For long-running (and typically daemonised) processes, it's industry norm to parameterise it with a configuration file rather than command line arguments especially if the program's configuration needs to be easily persisted through restarts. Complex configuration, and hot-loading configurations, maintainability and config modification also definitely benefit from this plumbing in place.

Worth taking some time to discuss this group-wise?

internal/femto/femto.go
Euphoride: That makes more sense, that's definitely worth looking into then. I'll make sure to fix that and have some interface for deserialisation that checks for these failures.

PR #37, opened 2024-01-23
Implement Postgres database instance

### LINKED ISSUE #13: Set up link to postgres db

danielg0:
Write an implementation of the Database interface that works with a postgres database via the sql module

Will need some CI changes in order to pass tests

Closes #13


danielg0:
This PR is ready to start being reviewed. Postgres implementation is now written and all that is left is:

- [x] Fixing/Writing unit tests
- [x] Tidying up the setup in `groundstation/main.go` (I've left the examples of usage in there as reference for the moment)
- [ ] Integrating it into the other parts of groundstation (should be done in other prs)

README.md
danielg0, line 1: Note this README is outdated and will be corrected as testing is being done.

TLDR is that production postgres db is taken from env var `DATABASE_URL`, and I'm planning on having the tests use a separate database taken from `TEST_URL` (so we can (accidentally or purposefully) run the tests on the production machine without risking it's data)

cmd/groundstation/main.go
```diff
@@ -20,7 +22,19 @@ func main() {
 
 	srv := groundstation.NewServer()
 
-	slog.Info("Stating groundstation API server", "port", configuration.Server.Port)
+	// open database connection
+	dbUrl := os.Getenv("DATABASE_URL")
```
Euphoride: this should come from the toml configuration file

cmd/groundstation/main.go
```diff
@@ -20,7 +22,19 @@ func main() {
...
+	db, err := postgres.New(dbUrl)
+	if err != nil {
+		slog.Error("when opening database", "err", err)
+		return
+	} else {
+		slog.Info("connected to database", "url", dbUrl)
+	}
+
+	_ = db
```
Euphoride: I think this line is unusual, what's the intention here?

internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,265 @@
...
+		// if not shut before the next transaction, causes
+		// "driver: bad connection" errors
+		err = rows.Close()
+		if err != nil {
+			return errors.Join(err, tx.Rollback())
+		}
+
+		now := time.Now()
+		if lastSeen.Before(now) {
+			slog.Debug("last seen was before now, update it")
```
Euphoride: Refactor this into it's own function for sake of modularity

internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,265 @@
...
+
+		// Next() failing might be because of an error
+		err = rows.Err()
+		if err != nil {
+			return errors.Join(err, tx.Rollback())
+		}
+
+		// a machine with this hostname wasn't found, so make a new row
+		// for it
+		_, err = tx.Exec(`INSERT INTO Machines (Hostname, LastSeen)
```
Euphoride: This too, this is a significant enough change in behaviour (i.e. transition from fail state to managed fail state) that it'd benefit from it's own name

internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+
+		// TODO: in future, it should be added to a list to wait for
+		// approval
+	}
+
+	return tx.Commit()
+}
+
+func (conn postgresConn) AppendDataPoint(host string, packet uplink.GPUStats) error {
```
Euphoride, line 151: Similar comment for the structure of this function - would benefit from being the composition of multiple. Makes testing admittedly more tedious, but I think it would be worth it.

internal/database/postgres/postgres.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	return tx.Commit()
+}
+
+func (conn postgresConn) LatestData() (map[string][]uplink.GPUStats, error) {
+	rows, err := conn.db.Query(`SELECT g.Machine, g.Name, g.Brand,
+			g.DriverVersion, g.MemoryTotal, s.MemoryUtilisation,
+			s.GpuUtilisation, s.MemoryUsed, s.FanSpeed, s.Temp
+		FROM GPUs g INNER JOIN Stats s ON g.Id = s.Gpu
```
Euphoride, line 225: Trying to follow this query - are three inner joins necessary?

internal/database/postgres/postgres_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+// run all the database unit tests on the postgres implementation
+func TestPostgres(t *testing.T) {
+	// set default value that matches github workflow
+	url := os.Getenv("TEST_URL")
+	if url == "" {
+		url = "postgres://postgres@localhost/postgres"
+	}
+
+	for _, test := range database.UnitTests {
```
Euphoride, line 18: Actually I would suggest that you wrap the entire test in a transaction rather than create/drop tables.

internal/database/unit_tests.go
```diff
@@ -0,0 +1,127 @@
...
+	{"AppendedDataPointsAreSaved", appendedDataPointsAreSaved},
+	{"MultipleHeartbeats", multipleHeartbeats},
+}
+
+// fake data for adding during tests
+var fakeData = uplink.GPUStats{Name: "GT 1030", Brand: "NVidia",
+	DriverVersion: "v1.4.5", MemoryTotal: 4, MemoryUtilisation: 25.4,
+	GPUUtilisation: 63.5, MemoryUsed: 1.24, FanSpeed: 35.2, Temp: 54.3}
+
+// functions for approximately comparing floats and data structs
```
Euphoride, line 33: Can we directly override float equality? Unsure if this is possible in Go.

Euphoride (commented):
Some general comments and questions, happy with semantics/correctness and just reviewing style mainly.

cmd/groundstation/main.go
danielg0: Yep, can do. Should a make a new category for database settings?

cmd/groundstation/main.go
danielg0: Go raises errors on variables that aren't used - this line uses the variable for nothing to make the error go away. Once the database has been integrated, the line can be deleted

cmd/groundstation/main.go
Euphoride: Yes, I reckon that should be the cleanest way

cmd/groundstation/main.go
Euphoride: Ahh that makes sense. Would it be worth `_`'ing the db variable as soon as it'd defined and noting this in a comment?

internal/database/unit_tests.go
danielg0: I don't believe so. Tbh, these functions are a temporary fix - I've left a todo at the top of the file to remind myself/someone else to move them elsewhere and define some more utilities for testing (ExpectEqual, ExpectToFail, Try, etc.)

cmd/groundstation/main.go
danielg0: You're quite right, I'm used to zig where that's not allowed. I'll make that change in a moment

internal/database/postgres/postgres.go
danielg0: I think so?

We want to get all the associated info for the latest entry in `Stats` (ie. the ones with the max recieved) so we:
- do a sub-query that aggregates every Stat row to get the largest recieved timestamp per stats.Gpu
- join this back with the Stats table to get the "dynamic" info for that row (ie. gpu usage, fan speed, etc.)
  - I don't believe we can do this in the aggregating sub-query because we don't want to aggregate gpu usage/fan speed, we just want the exact values from the row that's latest
 - then inner join that with the gpus table to get the "static" info (ie. driver version, machine hostname, etc.)

danielg0: I could add that explanation and move the sub-query so it's the first thing taken from to better illustrate that journey if that'd help?

internal/database/postgres/postgres_test.go
danielg0: I don't think that would be doable (at least working within the confines of the go sql library). The postgres implementation already wraps most of its work in transactions, so executing tests within a transaction would require nested transactions, and the transaction type [`sql.Tx`](https://pkg.go.dev/database/sql#Tx) doesn't have a function for creating transactions within it.

In theory nested transactions look like they'd be possible using [savepoints](https://stackoverflow.com/questions/25425944/does-postgres-support-nested-or-autonomous-transactions), but I'm not sure how we could elegantly use savepoints whilst testing, but then go back to using transactions during production?

danielg0: It would be great if possible though, because then we could use Postgres' temporary tables (tables that're only visible within one transaction and destroyed after) and avoid a whole host of issues that currently can occur if you run `go test ./...` in more than one terminal at a time

internal/database/postgres/postgres.go
danielg0: Sure, something like `createMachine(hostname string)`?

internal/database/postgres/postgres.go
danielg0: Do you specifically mean pulling out getting the last seen time into it's own function?

cmd/groundstation/main.go
danielg0: Done in 6133a1806b0ddb508ac84cbe6d6bfe1c5efe16e1

cmd/groundstation/main.go
danielg0: Done in 1ace56bd3f0bcf687400a503dbe393c26c729570

internal/database/unit_tests.go
danielg0: Looks like I can use testify for that

internal/database/postgres/postgres.go
Euphoride: Yes I'd agree with that

internal/database/postgres/postgres.go
Euphoride: I might suggest this:

```sql
SELECT 
    g.Machine, 
    g.Name, 
    g.Brand,
    g.DriverVersion, 
    g.MemoryTotal, 
    s.MemoryUtilisation,
    s.GpuUtilisation, 
    s.MemoryUsed, 
    s.FanSpeed, 
    s.Temp
FROM GPUs g
INNER JOIN (
    SELECT *,
           ROW_NUMBER() OVER(PARTITION BY Gpu ORDER BY Recieved DESC) as rn
    FROM Stats
) s ON g.Id = s.Gpu
WHERE s.rn = 1
```

internal/database/postgres/postgres_test.go
Euphoride: Ah that's annoying, I suppose we'll have to keep on the look out for that breaking then.

Issue #36, opened 2024-01-23
Send GPU data packets from satellite to groundstation


YRMYJASKA:
Add satellite-side code that will use the groundstation submission API to submit data about GPUs

PR #39, opened 2024-01-23
Add satellite -> groundstation messaging

YRMYJASKA:
Simple enough. Just added so we can have some way of testing the whole pipeline (satellite -> groundstation -> api -> user) soon.

### DIFF
--- a/cmd/groundstation/main.go
+++ b/cmd/groundstation/main.go
@@ -14,7 +14,6 @@ import (
 )
 
 func main() {
-	slog.Info("Starting groundstation")
 
 	configuration, err := config.GetConfiguration("config.toml")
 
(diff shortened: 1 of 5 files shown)


YRMYJASKA:
Closes #36

cmd/groundstation/main.go
```diff
@@ -14,7 +14,6 @@ import (
 )
 
 func main() {
-	slog.Info("Starting groundstation")
```
aDotInTheVoid, line 17: NIT: Why'd this go? I'd like to keep it before we load any config, so you can see that it's started loading the config.

YRMYJASKA: No reason in particular, just shuffled things because I wanted to see "started groundstation at port xxxx" but I have no strong feelings about either way.

PR #40, opened 2024-01-23
cmd/groundstation: Use femto for status submission.

aDotInTheVoid:
This allows removing JSON and HTTP method logic from the `HandleStatusSubmission` function.

It also means all the methods are on the same mux.

Doing this requires moving the API stuff into it's own library, so it can be tested nicely. We'll also want this for when the web api, and ground API are on the same binary.


aDotInTheVoid:
saw aproval from mila on discord

PR #43, opened 2024-01-23
all: Reorganize backend

aDotInTheVoid:
- Remove `incrementer`
- status: Merge `status/handlers` into `status`
- Rename `status` to `gpustats`
- Rename `status.GPUStatusPacket` to `uplink.GPUStats`
- groundstation: Move most of implementation to `internal/groundstation`.


aDotInTheVoid:
I've not moved `database` or `config` yet, to not disrupt #37, but they can be done later.

danielg0:
Right okay, so let me check my understanding of what's going on here:

- We have a `cmd` folder that contains folders for each binary we're producing - they only contain the `main.go` file required for building.
- Every other module is in `internal` (because we don't produce libraries) and is either a sub-module of `satellite`/`groundstation` when the behaviour is specific to that binary or just a sub-module of `internal` when behaviour is shared between them

If so, that looks good to me. I'll rebase #37 and move the `database` module to `internal/groundstation/database`

danielg0 (approved):

aDotInTheVoid:
Almost:

Go doesn't have a concept of "sub-module"s or sub-packages. IE theirs no relationship between `internal/groundstation` and `internal/groundstation/database`, other than having similar paths. Also the name of the packages would be `groundstation` and `database`.  (This is why in std, it's `"net/http/httptest"`, not `"net/http/test"`.

Therefor, I'm aiming for a much flatter package layout. IE the DB should live in `internal/database`. (Also, it'd also be used by webapi, so it's not groundstation specific). This also means theirs little point in having `internal/database/postgres` imo.

See <https://go.dev/doc/modules/layout#server-project> for the layout we're going for.

(aside: your using module where you mean package. Everything in this repo is in one module, but it's split into packages. Modules are used for versioning, but packages are used for visibility, coherency and separate compilation. See <https://go.dev/ref/mod#modules-overview> for details).

PR #48, opened 2024-01-26
Annotating parallel on tests

Euphoride:
Simple PR! Annotating tests with `t.Parallel()` to improve execution speeds.


aDotInTheVoid:
Followup that i'm not sure is worth it: Enforce this in CI with <https://github.com/kunwardeep/paralleltest>.

At work, we had a quite nice convention where tests were enforced to have `t.Parallel()`, but those that couldn't used `//nolint:paralleltest`, which by convention also contained the explanation as to why.

PR #51, opened 2024-01-26
webapi: Use `femto`

aDotInTheVoid:
This moves the webapi into it's own module implementing `http.Handler`. It also uses the `femto` framework to abstract away routing and JSON reponces.

Makes no change to the logic itself, but makes it possible to implement those.


danielg0 (approved):
LGTM

PR #52, opened 2024-01-26
groundstation DB integration

danielg0:


cmd/groundstation/main.go
```diff
@@ -10,29 +10,28 @@ import (
 )
 
 func main() {
-	slog.Info("Starting groundstation")
```
aDotInTheVoid, line 13: NIT: Theirs no reason to remove it. I'd like to have it log immediatly

cmd/groundstation/main.go
```diff
@@ -10,29 +10,28 @@ import (
...
 	if err != nil {
 		slog.Error("when opening database", "err", err)
 		return
 	} else {
 		slog.Info("connected to database", "url", dbUrl)
 	}
 
-	slog.Info("Starting groundstation API server", "port", configuration.Server.Port)
+	// start servers
+	slog.Info("Starting groundstation dish server")
```
aDotInTheVoid, line 34: NIT: wait, where did this terminology come from

aDotInTheVoid (approved):

aDotInTheVoid:
If you feel like fixing the nit's that'd be nice, but it's late, so feel free to just not honestly.

cmd/groundstation/main.go
danielg0: Ah, I thought this message was indicating that the satellite->groundstation server was about to be started, so I moved it down to just before that actually occured, like how there's one for the api server

aDotInTheVoid: No, it's that the *binary* has started. Their can be a seperate line for the server

danielg0: Cool, will add back

cmd/groundstation/main.go
danielg0: The API server had a name (api) and a message indicating it had booted, so I thought the satellite->groundstation could use one too.
Happy to take suggestions on what we call the communication from satellite->groundstation, but I personally do quite like dish (it's like the reciever/transmitter to satellites on a groundstation)

danielg0: Alternatively, we could make the "groundstation api server", just "api server" and make the satellite->groundstation server the "groundstation server"

aDotInTheVoid: Can we bikeshed this another day please!!

danielg0: Of course. I'm sure we'll have a wheely great time

PR #55, opened 2024-01-26
webapi: Pull data from database, run in same binary as goundstation

aDotInTheVoid:
Still needs to create the DB


aDotInTheVoid:
I've switched away from toml config, as it's a faff for local development, where you don't want to have to edit a file to change settings.

cmd/webapi/main.go
```diff
@@ -10,7 +10,8 @@ import (
 func main() {
```
danielg0: What's the aim of keeping a main for webapi now's it's been merged into control?

And won't it crash on it's first request due to the db being nil?

aDotInTheVoid: Oh, yeah, this can be deleted (groundstation too)

danielg0 (approved):
I'm fine with this PR, but I think we need to decide on a configuration format and stick to it. We now have:
- An unused toml implementation in `cmd/groundstation/config`
- Postgres unit tests that only use environment variables and fallback to a hardcoded value
- control.go that uses go flags and an environment variable that fallsback to the postgres default server

danielg0:
To make myself clear, this is a topic for a discussion later, not now

PR #56, opened 2024-01-26
satellite: Add option to produce fake data.

aDotInTheVoid:
This means we can run locally, without needing a NVidia GPU.


danielg0 (approved):
LGTM

PR #60, opened 2024-01-26
Added npm fmt CI step

Euphoride:


frontend/package.json
```diff
@@ -27,7 +27,8 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "fmt": "prettier --check ."
```
aDotInTheVoid: NIT: I'd love this to be called `fmt-check`, to be clear it won't format

cmd/groundstation/main.go
```diff
@@ -0,0 +1,28 @@
+package main
```
aDotInTheVoid: Why are we bringing this back? It was intentionally deleted

frontend/package.json
Euphoride: Yes, I think that's ideal too

cmd/groundstation/main.go
Euphoride: I think switching from an old branch caused some issues, I'll be fixing this soon. Should have labelled the PR draft :P

aDotInTheVoid:
r=me with commits squashed (if you know how to do that, otherwise as-is is fine)

PR #62, opened 2024-01-26
Write Dockerfiles for deploying to VM

### LINKED ISSUE #79: Coalesce on TOML configuration
Replace all uses of environment variables and command-line flags with config from the TOML file

danielg0:
Closes #79


aDotInTheVoid:
Why are we having the frontend and control server be in docker images?

I thought the point was to use caddy to run them on the same machine, so it's easy to deploy.

danielg0:
> I thought the point was to use caddy to run them on the same machine, so it's easy to deploy.

I agree, that's why I've done them as Dockerfiles, with the frontend being a Caddy container that proxies to Control. All the containers will run on the same machine, no dependencies needed beyond docker & docker compose. Then, compose.yaml will join them together so deploying is as simple as:

`sudo docker compose up --build -d`

aDotInTheVoid:
Fine for now, but i’d like to take a crack at a single image deployment later, as i think that’s be nicer than requiring compose

danielg0:
This is done & ready for review, insofar as we can deploy somewhere, and I don't want to cram anymore stuff into this PR. Things still left (for other prs):

- [ ] HTTPS
  - Using DNS api from Hetzner 
  - Will require us to start storing secrets on deployment
- [ ] Move postgres to use passwords
  - Requires previously mentioned secret management
- [ ] Clean database
  - ie. regularly remove stats from the db that are old
  - have some algorithm for keeping/averaging to 1 stat for each day so we can do long-term graphing/analysis
- [ ] Add CI pipeline testing the the dockerfiles build

README.md
```diff
@@ -2,8 +2,8 @@
 
 ## Deploying
 
-Needs a Postgres database. URL is passed in as `DATABASE_URL`. For _reasons_,
-don't use `postgres://postgres@localhost/postgres`.
+Needs a Postgres database. URL is passed in in the `control.yaml` file. For
```
aDotInTheVoid: This should probably be either `control.toml` or `compose.yaml`

cmd/control/control.go
```diff
@@ -1,67 +1,61 @@
...
-	inMemDb := flag.Bool("inmemdb", false, "Use an transient, in-memory database")
-	postgresDb := flag.Bool("postgres", false, "Use a posgresql database from GPU_DB_URL")
-
-	flag.Parse()
-
 	log := slog.Default()
 	log.Info("Starting control server")
 
-	db, err := initialiseDatabase(inMemDb, postgresDb)
+	conf, err := config.GetServerConfiguration("control.conf")
```
aDotInTheVoid: Why `.conf` and not `.toml`? The latter is clearer

control.conf
```diff
@@ -0,0 +1,7 @@
+[Server]
```
aDotInTheVoid: This should be `control.toml.example`, and `control.toml` should be `.gitignore`d, so we change change our local config, without the risk of it showing up in git (this trick taken from rustc's build system)

frontend/src/App.tsx
```diff
@@ -15,7 +15,7 @@ import { Navbar } from "./Components/Navbar";
 import { ColumnGrid } from "./Components/ColumnGrid";
 import { WorkStationGroup } from "./Data";
 
-const API_URL = "http://localhost:8000";
+const API_URL = "";
```
aDotInTheVoid, line 18: Probably worth adding a comment as to why this works in prod?

Also, does this mean the frontend only works when on the same port as the webapi? That seems quite bad for development.

internal/config/server_configuration.go
```diff
@@ -1,11 +1,16 @@
...
+	WAPort int `toml:"webapi_port"`
 }
 
 type Database struct {
-	Url string `toml:"url"`
+	InMemory    bool   `toml:"inmemory"`
+	Postgres    bool   `toml:"postgres"`
+	PostgresUrl string `toml:"url"`
+	// TODO: this is currently unused - need to figure out how to pass toml config to tests
+	TestUrl string `toml:"test_url"`
```
aDotInTheVoid: Tests shouldn't load from toml IMO.

internal/config/client_configuration.go
```diff
@@ -49,7 +50,7 @@ func (s SatelliteConfiguration) Merge(config Mergable) Mergable {
 func DefaultSatelliteConfiguration() SatelliteConfiguration {
 	return SatelliteConfiguration{
 		Groundstation: Groundstation{"localhost", 8080},
-		Satellite:     Satellite{"/tmp/satellite", 60, 2},
+		Satellite:     Satellite{"/tmp/satellite", 60, 2, false},
```
aDotInTheVoid: NIT: At this point, this should use field names, as it's not at all obvious what's going on here

internal/config/server_configuration_test.go
```diff
@@ -52,8 +59,10 @@ server: "should be a table, not a string"`
 
 	config, err := config.GetServerConfiguration(filename)
 	assert.Error(t, err)
-	assert.Equal(t, 0, config.Server.Port)
-	assert.Equal(t, "", config.Database.Url)
+	assert.Equal(t, 0, config.Server.GSPort)
```
aDotInTheVoid: NIT: This can probably be `assert.Equal(t, ServerConfig{}, config)`

satellite.conf
```diff
@@ -0,0 +1,8 @@
+[Groundstation]
```
aDotInTheVoid: DITTO re naming

cmd/satellite/main.go
```diff
@@ -29,7 +26,9 @@ func main() {
 
 	log.Info("got hostname", "hostname", host)
 
-	satellite_configuration, err := config.GetClientConfiguration("config.toml")
+	satellite_configuration, err := config.GetClientConfiguration("satellite.conf")
```
aDotInTheVoid: DITTO re extension

frontend/src/App.tsx
Euphoride: This seems like it actually shouldn't be possible. Static serving and the webapi shouldn't be able to listen on the same port no matter how hard we try. I think we definitely need to be aware of this one.

internal/config/server_configuration.go
Euphoride: I'm happy for tests to read from toml if it's the most efficient way to set up the state for the testable code, but there is always the option of mocking whatever state instead. I think we should go for the latter option too.

internal/config/server_configuration.go
```diff
@@ -20,21 +25,35 @@ func (s ControlConfiguration) Merge(config Mergable) Mergable {
 		return s
 	}
 
-	if s.Database.Url == "" {
-		s.Database.Url = sat_config.Database.Url
+	if s.Database.PostgresUrl == "" {
```
Euphoride, line 26: Just a general comment but this pattern would be ideally abstracted somehow. Is Go capable of reflection on it's structs?

internal/config/server_configuration_test.go
```diff
@@ -12,19 +12,26 @@ func TestGetConfiguration_ValidConfig(t *testing.T) {
...
 [server]
-port = 9090
+groundstation_port = 9090
+webapi_port = 9070
 
 [database]
-url = "postgres://tony@ic.ac.uk/squares"`
+postgres = true
+url = "postgres://tony@ic.ac.uk/squares"
+test_url = "postgres://postgres@localhost/gpuctl-tests"`
```
Euphoride, line 21: Ahh, so this is the testing environment state referred to earlier. I believe this should be an immutable constant in our test code, or a dev environment variable. Either of these have their own issues too, so if we are prioritising fast test configuration changes then maybe we should separate out the testing toml config into it's own file.

internal/config/server_configuration.go
aDotInTheVoid: yeah, but it’s a faff, and not worthwhile here (as this code’s going to be removed anyway).

if we did need to do this, using generate would probably be nicer than reflection, but that’s just speculation on my part

frontend/src/App.tsx
aDotInTheVoid: this works via caddy and reverse proxy. that’s kind of the whole point

Euphoride: Ah, missed that. Now that I’ve had some time to contemplate it, I recall quite a few sites having /api/ routes that get hit on the same port as the static server (via reverse proxy) and it’s quite an elegant usage pattern. 

I’d argue here in that case that it’s nicer this way then.

README.md
danielg0: Yeah, sleepy misspelling on my part, will fix soon

cmd/control/control.go
danielg0: That is true. It started as one `config.toml` which needed to be two, so I pushed the `config` down into the extension, but `control.toml` is probably distinguishing enough

control.conf
danielg0: Sure

frontend/src/App.tsx
danielg0: I can add a comment, but I'd prefer to keep it like this. Setting up a dev environment should be easier now we have docker compose, making it configurable gives us another thing that can break even when it works in dev, and if our api routes are on the same endpoint, I believe that means we can restrict our cors rules a lot more (at least, the only reason we have them disabled currently is because the dev server on localhost:3000 complained about communicating with localhost:8080).

We can probably make our URIs more elagant once this is by renaming `/gs-api` to `/groundstation` and `/api` to something more meaningful

danielg0: Or alternatively `/api/groundstation` and `/api/web`?

internal/config/server_configuration.go
danielg0: Yeah, I can go back and remove this because I don't think tests with TOML is doable due to them all being run in temp directories containing just the binary being tested

internal/config/server_configuration_test.go
danielg0: Fixed in 1229c0e

internal/config/server_configuration.go
danielg0: Removed in 1229c0e

README.md
danielg0: Fixed in 7e12fd7

cmd/control/control.go
danielg0: Fixed in 7e12fd7

control.conf
danielg0: Fixed in 7e12fd7

README.md
danielg0: Fixed in 762600c

cmd/control/control.go
danielg0: Fixed in 762600c

control.conf
danielg0: Fixed in 762600c

satellite.conf
danielg0: Fixed in 762600c

cmd/satellite/main.go
danielg0: Fixed in 762600c

internal/config/client_configuration.go
danielg0: Fixed in 52ff904

internal/config/server_configuration_test.go
danielg0: Fixed in 52268e8

PR #68, opened 2024-01-29
Refactored configuration code + Added client configuration code

Euphoride:
See title. PR is welcome to reviews.


Euphoride:
(This probably shouldn't be merged until I can write proper tests for it. Or, optionally, we merge it and I start a new PR to improve test coverage)

internal/groundstation/config/client_configuration.go
Euphoride, line 1: This really should be moved to a different path that isn't `internal/groundstation`. Noting to make it clear I am aware.

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
 	log.Info("got hostname", "hostname", host)
 
+	satellite_configuration, err := config.GetClientConfiguration("config.toml")
+
+	if err != nil {
+		log.Error("Failed to get satellite configuration from toml configuration file", "err", err)
+	}
+
 	s := satellite{
 		// TODO: Make this configurable
```
danielg0: Can this TODO line be removed now hostname is configurable?

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
+	go func() {
+		for {
+			log.Info("Sending heartbeat")
+			err := s.sendHeartBeat()
+
+			if err != nil {
+				log.Error("failed to send heartbeat", "err", err)
+			}
+
+			// TODO: testing only, should not send packets this frequently?
```
danielg0: Same for this TODO

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
-	for {
-		log.Info("Sending heartbeat")
-		err := s.sendHeartBeat()
-		if err != nil {
-			log.Error("failed to send heartbeat", "err", err)
+		for stat := range backlog {
+			err := s.sendGPUStatus(backlog[stat])
+
+			if err != nil {
+				log.Error("failed to send status", "err", err)
```
danielg0: Would it be possible to distinguish this logging error message to specify it's a message from the backlog? At the moment, it's the same error as if the backlog was empty and the first message sent failed.

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
+				if err != nil {
+					log.Error("failed to send status", "err", err)
+				}
+			case <-collectGPUStatTicker.C:
+				log.Info("Collecting GPU Status")
+
+				stat, err := hndlr.GPUStats()
+
+				if err != nil {
+					log.Error("Failed to get GPU stat from stat handler")
```
danielg0: Can we log the particular error that occurred here too?

danielg0 (approved):
LGTM, other than a couple minor logging/outdated comments.

I'm right in thinking that we're currently only sending the most recent stats in the backlog every time we publish a set of gpu stats, and the plan is to add aggregation in a later PR?

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
 		log.Error("failed to get hostname", "err", err)
 		return
 	}
+
 	log.Info("got hostname", "hostname", host)
 
+	satellite_configuration, err := config.GetClientConfiguration("config.toml")
+
+	if err != nil {
+		log.Error("Failed to get satellite configuration from toml configuration file", "err", err)
```
aDotInTheVoid, line 35: `Logger.Error` just logs, but doesn't exit. <https://pkg.go.dev/log/slog#Logger.Error>. If we fail to load, we should also exit the program

cmd/satellite/main.go
```diff
@@ -1,63 +1,105 @@
...
 	host, err := os.Hostname()
+
 	if err != nil {
 		log.Error("failed to get hostname", "err", err)
 		return
 	}
+
 	log.Info("got hostname", "hostname", host)
 
+	satellite_configuration, err := config.GetClientConfiguration("config.toml")
```
aDotInTheVoid, line 32: NIT: This variable name is uncomfortably long.

cmd/satellite/main.go
```diff
@@ -67,20 +109,79 @@ type satellite struct {
...
+}
+
+func saveState(filename string, items []uplink.GPUStats) error {
+	data, err := json.Marshal(items)
+
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(filename, data, 0644)
```
aDotInTheVoid: NIT: This should be `0o644`, to be clearer that it's octal

cmd/satellite/main.go
```diff
@@ -67,20 +109,79 @@ type satellite struct {
...
+func parseProgramFlags() flags {
+	fakeGpus := flag.Bool("fakegpu", false, "Use fake GPU data")
+	flag.Parse()
+
+	return flags{
+		fakeGPUs: *fakeGpus,
+	}
+}
+
+func setGPUHandler(isFakeGPUs bool) gpustats.GPUDataSource {
```
aDotInTheVoid, line 156: Why's this named this? `setGPUHandler` makes it sound like it's a setter for some state, when it isn't. How about `makeGpuHandler`?

internal/config/client_configuration.go
```diff
@@ -0,0 +1,31 @@
+package config
+
+type ClientConfiguration struct {
+	Groundstation struct {
+		Hostname string `toml:"hostname"`
+		Port     int    `toml:"port"`
+	} `toml:"groundstation"`
+	Satellite struct {
```
aDotInTheVoid: NIT: These should probably be their own types, so that we don't need the tag data everywhere.

internal/config/client_configuration.go
```diff
@@ -0,0 +1,31 @@
+package config
+
+type ClientConfiguration struct {
```
aDotInTheVoid: This should be `SataliteConfiguration`, as "client" is super ambiguous.

internal/config/common.go
```diff
@@ -0,0 +1,84 @@
...
+
+func GetConfiguration[T any](filename string, defaultGenerator func() T) (T, error) {
+	exePath, err := os.Executable()
+
+	if err != nil {
+		return defaultGenerator(), err
+	}
+
+	exeDir := filepath.Dir(exePath)
+	configPath := filepath.Join(exeDir, filename)
```
aDotInTheVoid: This seems like a bad way to find the config file. Eg `go run ./cmd/satalite` will have the executable inside `/tmp/go-build/...`, and we don't want to need to place an config file their

internal/config/common.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if fileEmpty {
+		return defaultGenerator(), nil
+	}
+
+	var config T
+
+	if _, err := toml.DecodeFile(configPath, &config); err != nil {
+		return defaultGenerator(), nil
```
aDotInTheVoid: You probably meant to return the error here

internal/config/common.go
```diff
@@ -0,0 +1,84 @@
...
+	configPath := filepath.Join(exeDir, filename)
+
+	fileEmpty, err := IsFileEmpty(configPath)
+
+	if err != nil {
+		return defaultGenerator(), err
+	}
+
+	if fileEmpty {
+		return defaultGenerator(), nil
```
aDotInTheVoid, line 50: These semantics are really weird. Why do we say

- No file -> Error
- Empty file -> default config

internal/config/common.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	var config T
```
aDotInTheVoid: Does this mean we'll use the zero-value for fields not set in the config file?

internal/config/common.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if _, err := toml.DecodeFile(configPath, &config); err != nil {
+		return defaultGenerator(), nil
+	}
+	return config, nil
+}
+
+func CreateTempConfigFile(content string, t *testing.T) (string, func()) {
```
aDotInTheVoid: This can live in `config_test` instead of `config`

internal/config/server_configuration.go
```diff
@@ -0,0 +1,25 @@
+package config
+
+type ServerConfiguration struct {
```
aDotInTheVoid: Server is ambiguous. I think you mean control

internal/config/server_configuration.go
```diff
@@ -0,0 +1,25 @@
...
+	Server struct {
```
aDotInTheVoid: NIT: These should be their own types

Euphoride:
> LGTM, other than a couple minor logging/outdated comments.
> 
> 
> 
> I'm right in thinking that we're currently only sending the most recent stats in the backlog every time we publish a set of gpu stats, and the plan is to add aggregation in a later PR?

Yes, at the moment the process is choosing the most recent but we should implement some sort of aggregation soon.

cmd/satellite/main.go
Euphoride: I refactored this out as it was setting state rather than something resembling a factory method. I think set might be more appropriate here.

cmd/satellite/main.go
Euphoride: Is our styling convention avoiding long variable names atm? I much prefer this due to code clarity.

internal/config/client_configuration.go
Euphoride: I'd rather keep code locality as much as possible on this. This was an explicit decision not to extract out types as they're not used elsewhere. I would start to get worried if the configuration bundle gets massive, but by then it belongs in it's own file.

internal/config/common.go
Euphoride: We absolutely shouldn't be running the binary in production via go run. This is more of a production oriented decision as I'd expect the binary and the configuration file to be localised.

internal/config/common.go
Euphoride: Yes I did, oops :)

internal/config/common.go
Euphoride: This is deliberate. I think a config file should always be present as part of deployment, but an empty one represents a default configuration. From the perspective of the eventual maintainer, these semantics feel intuitive.

internal/config/common.go
Euphoride: This is a good point. I'd like semantics where we use defaults instead. Will look into this.

internal/config/common.go
Euphoride: Nope :) Circular dependencies are a thing then.

internal/config/server_configuration.go
Euphoride: Mentioned/addressed earlier

internal/config/common.go
aDotInTheVoid: Yeah, but it means you're breaking `go run` for local dev, which super sucks, and I don't want to approve code that does that

internal/config/common.go
aDotInTheVoid: If we have the semantics of using the default (as opposed to zero) for unset fields, then this is fine, and will happen automaticly (instead of needing this check). I'd be fine with that.

internal/config/common.go
aDotInTheVoid: Why? it's only used in `config_test`

internal/config/client_configuration.go
aDotInTheVoid: But they are used twice (in definition, and in default func). I'd like to be able to pass around just one of them at some point. It also makes the default funcs quite unreadable

internal/config/common.go
Euphoride: Hmmm, I'll have to think about an elegant way to fix this.

internal/config/common.go
Euphoride: Ah you mean moving it as a helper function inside common_test? I'll get that sorted.

internal/config/client_configuration.go
Euphoride: That's fair. We can look into that then.

cmd/satellite/main.go
```diff
@@ -33,12 +33,11 @@ func main() {
 
 	if err != nil {
 		log.Error("Failed to get satellite configuration from toml configuration file", "err", err)
+		os.Exit(-1)
```
aDotInTheVoid: NIT: Go'd rather this was in [1, 125]: https://pkg.go.dev/os#Exit

internal/config/common.go
```diff
@@ -4,11 +4,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"testing"
 
 	"github.com/BurntSushi/toml"
 )
 
+type Configurable interface {
+	Merge(config Configurable) Configurable
```
aDotInTheVoid: NIT: I don't love this interface, but at least for now you should document the invariants (that seem to be the `config` should be the same type as the receiver, and this type will be returned).

Also a random nit about naming: This would seem to suggest that a configurable is an "object that can be configured", whereas it's actually an "object that represents a configuration". Maybe something like `Mergable`.

internal/config/common.go
```diff
@@ -50,35 +53,7 @@ func GetConfiguration[T any](filename string, defaultGenerator func() T) (T, err
...
-		t.Fatal(err)
-	}
-
-	if err := tmpfile.Close(); err != nil {
-		t.Fatal(err)
-	}
-
-	return tmpfile.Name(), func() {
-		os.Remove(tmpfile.Name())
+		return defaultGenerator(), err
```
aDotInTheVoid: This would rather be the zero value than the default, as it shouldn't be accessed.

Euphoride: Given it shouldn't be accessed I'm not sure it matters too much

internal/config/common.go
```diff
@@ -50,35 +53,7 @@ func GetConfiguration[T any](filename string, defaultGenerator func() T) (T, err
...
 	}
+	return config.Merge(defaultGenerator()).(T), nil
```
aDotInTheVoid: You can avoid needing to implement merging, due to how decoding goes into an existing struct: https://go.dev/play/p/dx-RXKitbbk

internal/config/common_test.go
```diff
@@ -85,3 +87,31 @@ func TestGenerateAddress(t *testing.T) {
 		})
 	}
 }
+
+func CreateTempConfigFile(content string, t *testing.T) (string, func()) {
```
aDotInTheVoid, line 92: NIT: This doesn't need to be exported

internal/config/common.go
Euphoride: Go is Go so there's no native way to express that invariant in the type system. I'll look into documenting it comment-wise.

internal/config/common.go
```diff
@@ -27,7 +30,7 @@ func IsFileEmpty(path string) (bool, error) {
 	return fileInfo.Size() == 0, nil
 }
 
-func GetConfiguration[T any](filename string, defaultGenerator func() T) (T, error) {
+func GetConfiguration[T Configurable](filename string, defaultGenerator func() T) (T, error) {
```
aDotInTheVoid: does this need to be public?

internal/config/common.go
Euphoride: This seems helpful. Would mean we can remove the merge-related code.

internal/config/common.go
aDotInTheVoid: yeah, probably should have annotated with `NIT`. The advantage is that if you do ignore the error, you'll end up with a nonsense config, so the bug is more likely to be surfaced

internal/config/common_test.go
Euphoride: It does, as it's used in `config_test`

Euphoride: Although, some reshuffling means we can move things into their natural place. Will sort this out.

PR #77, opened 2024-01-29
The great refactor: changing the data format

YRMYJASKA:
Change the GPU stats we send from satellites and separate between static and dynamic data. Requires a LOT of refactoring.


danielg0 (approved):

YRMYJASKA:
@danielg0 This should be ready for you to work on the DB refactoring.

cmd/satellite/main.go
```diff
@@ -43,23 +44,35 @@ func main() {
 		hndlr = gpustats.NvidiaGPUHandler{}
 	}
 
+	// Send initial infopacket of GPUInfo
+	s.sendGPUInfo(hndlr)
+
+	// Set up neat ctrl-c breaking
+	termsig := make(chan os.Signal, 1)
```
aDotInTheVoid: Why are we making this change here?!?

This should be a seperate PR

internal/gpustats/gpustats.go
```diff
@@ -6,72 +6,51 @@ import (
...
+var (
+	SampleAdditionError = errors.New("two packets with different contexts cannot be aggregated using Add, consider using UncontextualAdd")
+)
 
-// Combine two GPUStats instances into one.
-func Add(l, r uplink.GPUStats) (uplink.GPUStats, error) {
-	if r.Name != l.Name || r.Brand != l.Brand || r.DriverVersion != l.DriverVersion || r.MemoryTotal != l.MemoryTotal {
-		// TODO: Expose this error publicly.
-		return uplink.GPUStats{}, errors.New("two packets with different contexts cannot be aggregated using Add, consider using UncontextualAdd")
+// Combine two GPUStatSamples instances into one.
```
aDotInTheVoid, line 13: NIT: What, what are the sematics of this function suposed to be? When would it be used?

I'm pritty sure adding 2 70C gpu's should make a 140C gpu.

Probably best to remove at some point, as I don't think it get's used.

internal/gpustats/gpustats.go
```diff
@@ -6,72 +6,51 @@ import (
...
-}
-
 // Scale each value in s by scalar
-func Scale(s uplink.GPUStats, scalar float64) uplink.GPUStats {
-	return uplink.GPUStats{
-		Name:              s.Name,
-		Brand:             s.Brand,
-		DriverVersion:     s.DriverVersion,
-		MemoryTotal:       s.MemoryTotal,
+func Scale(s uplink.GPUStatSample, scalar float64) uplink.GPUStatSample {
```
aDotInTheVoid, line 36: NIT: Ditto

internal/gpustats/gpustats.go
```diff
@@ -6,72 +6,51 @@ import (
...
-		MemoryTotal:   memoryTotal,
-
-		MemoryUtilisation: 0,
-		GPUUtilisation:    0,
-		MemoryUsed:        0,
-		FanSpeed:          0,
-		Temp:              0,
-	}
+func Default(name string, brand string, driverVersion string, memoryTotal uint64) uplink.GPUStatSample {
+	return uplink.GPUStatSample{}
```
aDotInTheVoid: NIT: If we're not going to be using any of the args, we can just remove this function (and churn the tests)

internal/uplink/gpustats.go
```diff
@@ -3,22 +3,41 @@ package uplink
 // TODO: change to "/gs-api/gpu-stats"
 const GPUStatsUrl = "/gs-api/status/"
 
-type StatsPackage struct {
-	Hostname string   `json:"hostname"`
-	Stats    GPUStats `json:"stats"`
+type GpuStatsUpload struct {
+	Hostname string          `json:"hostname"`
```
aDotInTheVoid, line 7: NIT: Now that we don't share these with the frontend, we could just remove these json tags.

aDotInTheVoid (commented):
Seems generally good. I don't know enough about DB stuff to review that.

As always, `NIT: ` comments don't need to be fixed before merging, but I'd like to land the singnal handling stuff seperatly, as it deserves consideration, but I don't want to block this on that

cmd/satellite/main.go
YRMYJASKA: Got carried away. Just a small thing that I wanted, but can remove if you want to.

internal/gpustats/gpustats.go
YRMYJASKA: yeah, agreed. I was just mindlessly refactoring this bit. Will remove this func.

cmd/satellite/main.go
aDotInTheVoid: Yeah, let's please deal with this seperatly.

internal/uplink/gpustats.go
YRMYJASKA: Maybe, let's see how the refactors actually impact the front-end first.

danielg0:
Now this is at a point where it compiles, I think we should consider continuing to use what's currently marked as `OldGPUStatSample` for the frontend and altering the database interface to return `workStationData`, rather than GpuStatsUpload (which, just being two slices of info and stats, has a whole bunch of assumptions the current code depends on - different number of items in each list, items at the same index being different GPUs, whether or not one is missing depending on the type of request - making the safest way to write this to be a reimp of an inner-join).

Ideally, I think we'd remove GpuStatsUpload, but we'd need to find some way of carrying around a hostname with a request without needing it in the struct the GpuStatsProvider returns. Perhaps we could add it to the headers on the request?

Then the database consumes the info on the start-up context push, and the stats whenever they come in, then on request spits out an-already zipped and grouped array (or map?) of hostnames and/to `[]OldGPUStatSample`, which is passed through to the frontend on a `all` call. Then in future if we want to add a per-machine info endpoint, we can use the same `OldGPUStatSample`, just only return one rather than an array.

danielg0:
Also, I think it would give us the possibility to streamline some of the struct names.

Why do we have a `GPUStatSample` but the context is just `GPUInfo` (ie. why not `GPUStat`?)

YRMYJASKA:
> Also, I think it would give us the possibility to streamline some of the struct names.
> 
> Why do we have a `GPUStatSample` but the context is just `GPUInfo` (ie. why not `GPUStat`?)

these can be changed at will later. Feel free to invent a better name.

internal/gpustats/gpustats.go
danielg0: I believe @Euphoride added it back in (https://github.com/gpuctl/gpuctl/commit/325072c55246cb373d2a03f7f319d330193d05f4), so this isn't new, but my understanding is with the satellite locally storing a backlog of status packets and only sending only every so often (#68), we want a way to aggregate all the packets together (by summing so we can divide and average, etc.) into the one packet we then send off

PR #81, opened 2024-02-01
Prevent satellite from dying instantly

NathanielB123:
Go main threads don't wait for goroutines to finish, so we need to loop somewhere in the main thread, and the collect/publish stats loop seems as good a place as anywhere.

Also make the satellite collect and publish stats instantly instead of waiting for the first tick. This means we get to see that the satellite is working much faster for larger intervals.

### DIFF
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -63,50 +63,58 @@ func main() {
 		}
 	}()
 
-	go func() {
-		backlog, _ := recoverState(satellite_configuration.Satellite.Cache)
+	backlog, _ := recoverState(satellite_configuration.Satellite.Cache)
 
-		for stat := range backlog {
-			err := s.sendGPUStatus(backlog[stat])
+	for stat := range backlog {
+		err := s.sendGPUStatus(backlog[stat])
 
-			if err != nil {
-				log.Error("Failed to send backlogged GPU stat message", "err", err)
-			}
+		if err != nil {
+			log.Error("Failed to send backlogged GPU stat message", "err", err)
+		}
+	}
+
+	backlog = make([][]uplink.GPUStatSample, 0)
+
+	collectGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval))
+	publishGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval))
+
+	// Go has no API for a ticker with an instantaneous first tick (see
+	// https://github.com/golang/go/issues/17601) so we have to use a clunky
+	// work-around
+	publishGPUStats := func() {
+		log.Info("Sending status")
+
+		err = s.sendGPUStatus(processStats(backlog))
+
+		if err != nil {
+			log.Error("Failed to publish current GPU stat message", "err", err)
+		}
+	}
+
+	collectGPUStats := func() {
+		log.Info("Collecting GPU Status")
+
+		stat, err := hndlr.GetGPUStatus()
+
+		if err != nil {
+			log.Error("Failed to get GPU stat from stat handler", "err", err)
 		}
 
-		backlog = make([][]uplink.GPUStatSample, 0)
+		backlog = append(backlog, stat)
+		saveState(satellite_configuration.Satellite.Cache, backlog)
+	}
 
-		collectGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval) * time.Second)
-		publishGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval) * time.Second)
+	collectGPUStats()
+	publishGPUStats()
 
-		for {
-			select {
-			case <-publishGPUStatTicker.C:
-
-				log.Info("Sending status")
-
-				err = s.sendGPUStatus(processStats(backlog))
-
-				if err != nil {
-					log.Error("Failed to publish current GPU stat message", "err", err)
-				}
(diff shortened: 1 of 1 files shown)


danielg0:
Great, thanks for fixing this Nathaniel, I was just about to have to do it myself

cmd/satellite/main.go
```diff
@@ -63,50 +63,58 @@ func main() {
...
+			log.Error("Failed to send backlogged GPU stat message", "err", err)
 		}
+	}
 
-		backlog = make([][]uplink.GPUStatSample, 0)
+	backlog = make([][]uplink.GPUStatSample, 0)
 
-		collectGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval) * time.Second)
-		publishGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval) * time.Second)
+	collectGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval))
```
danielg0, line 78: If we remove `* time.Second` here, does that mean that our configured satellite data & heartbeat intervals have to be in milliseconds? I think this is more precise than we need to be (ie. I'd rather change the heartbeat interval to be in seconds rather than change the tickers to not be in seconds)

cmd/satellite/main.go
```diff
@@ -63,50 +63,58 @@ func main() {
...
+	publishGPUStatTicker := time.NewTicker(time.Duration(satellite_configuration.Satellite.DataInterval))
 
-		for {
-			select {
-			case <-publishGPUStatTicker.C:
+	// Go has no API for a ticker with an instantaneous first tick (see
+	// https://github.com/golang/go/issues/17601) so we have to use a clunky
+	// work-around
+	publishGPUStats := func() {
```
danielg0, line 84: Do these need to be anonymous functions or could they be moved to be freestanding functions of their own?

aDotInTheVoid:
Kinda spooky that CI didn't catch this

cmd/satellite/main.go
```diff
@@ -63,50 +63,58 @@ func main() {
 		}
 	}()
 
-	go func() {
-		backlog, _ := recoverState(satellite_configuration.Satellite.Cache)
+	backlog, _ := recoverState(satellite_configuration.Satellite.Cache)
```
danielg0, line 66: I know it was @Euphoride that originally added the recovery, but I wanted to take the opportunity to ask why we're recovering the backlog on startup? I get the idea of having a backlog we fill up, aggregate and then send off, but why do we want to pick up the backlog from the last time the satellite ran. That data could be outdated, and as we take a timestamp on entry to the DB, the stats would be incorrectly treated as "new", right?

Not worth blocking the PR over though

danielg0 (changes requested):

cmd/satellite/main.go
aDotInTheVoid: Yeah, that's a seperate problem, I'm working on it

danielg0:
> Kinda spooky that CI didn't catch this

Because we don't have any integration tests in CI. All the component parts of the satellite work, just not the main function. Could be useful to have if someone wants to work on them (I can't rn)

aDotInTheVoid:
Yeah, I know why it's happening. I'm just commenting that I'm unsettled by the fact that this was broken, and only uncovered with manual testing. I don't think we can do CD until we make sure basic stuff like this is tested

cmd/satellite/main.go
aDotInTheVoid: They capture a load of state, so it'd be a bit of a pain

NathanielB123: They aren't going to be called from anywhere else, so I kinda think anonymous functions are neater

danielg0:
> Yeah, I know why it's happening. I'm just commenting that I'm unsettled by the fact that this was broken, and only uncovered with manual testing. I don't think we can do CD until we make sure basic stuff like this is tested

Yeah, the more time I've spent on Docker stuff, the more I think whilst we can make deployment _easy_, I don't think it can ever be automated without substantial shenanigans. Notably, how do we give a GitHub action access to a DoC machine and how do database migrations get automated (the gs rn really sucks at even noticing the table layout is wrong)

cmd/satellite/main.go
NathanielB123: The timestamp coming from the time the data is added to the DB seems like the root problem here to me. Is there a reason we don't timestamp the data packets?

danielg0 (approved):

cmd/satellite/main.go
danielg0: Because the satellites are all running on different machines, that might have different/out-of-sync clocks. By only taking the timestamp on entry in the db, we have one source of truth for timing.

Note for this same reason once we start displaying last seen times in the UI, I think the api endpoint should return the duration to when the last heartbeat occured, not the timestamp it occured on

Euphoride: Backlog is there to as much as possible help there be as close as possible to no blindspots in our data. The idea is, if we fail between data uploads, and recover, then we can assume either one of two semantics:

1. Assume all data within the intervals was uncollectible, start streaming data from the present
2. Attempt to recover our state as much as possible, upload backlog, restart timer and stream from present.

Our state is actually quite small in the satellite, easily recoverable and therefore if we can help it, I think we'd like to ensure the groundstation has close enough to perfect data.

I'd think of it as a QoL feature, rather than an critical feature, but worth keeping.

PR #85, opened 2024-02-01
Remove `act` call from JarJar hook

NathanielB123:
We should only ever call `act` inside tests (it is unsupported in production builds)
Unfortunately, the way I am using the state hook makes this quite challenging, so we do a janky mix of awaiting `waitFor` and `findBy`s to get Jest to stop complaining

### DIFF
--- a/frontend/src/App.test.tsx
+++ b/frontend/src/App.test.tsx
@@ -55,7 +55,7 @@ test("retrieves data from API server and displays correctly", async () => {
 
         screen
           .getAllByText(gpu.gpu_temp + " °C", { exact: false })
-          .forEach((temp) => expect(temp).toBeInTheDocument());
+          .forEach((gpu_temp) => expect(gpu_temp).toBeInTheDocument());
       });
     });
   });
@@ -73,7 +73,12 @@ test("data is fetched again after refresh interval", async () => {
   (await screen.findAllByText("31 °C", { exact: false })).forEach((temp) =>
     expect(temp).toBeInTheDocument(),
   );
-  data[0].workStations[0].gpus[0].gpu_temp = 100;
+
+  const group = data[0];
+  const workStation = group.workStations[0];
+  const gpu = workStation.gpus[0];
+  gpu.gpu_temp = 100;
+
   jest.advanceTimersByTime(REFRESH_INTERVAL + 1);
 
   await waitFor(() => expect(fetch).toHaveBeenCalledTimes(1));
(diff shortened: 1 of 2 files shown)


NathanielB123:
Huh, requiring trailing commas on function arguments is a slightly weird formatting requirement

aDotInTheVoid:
@Euphoride can you review this? I don't think anyone else if framilar enough with react to know what's going on here.

frontend/src/App.test.tsx
```diff
@@ -62,9 +70,9 @@ test("data is fetched again after refresh interval", async () => {
 
   await waitFor(() => expect(fetch).toHaveBeenCalledTimes(1));
 
-  screen
-    .getAllByText("31 °C", { exact: false })
-    .forEach((temp) => expect(temp).toBeInTheDocument());
+  (await screen.findAllByText("31 °C", { exact: false })).forEach((temp) =>
```
Euphoride, line 73: NIT: Might be worth having a more descriptive name here than `temp`

frontend/src/App.test.tsx
```diff
@@ -62,9 +70,9 @@ test("data is fetched again after refresh interval", async () => {
...
+    expect(temp).toBeInTheDocument(),
+  );
   data[0].workStations[0].gpus[0].gpu_temp = 100;
```
Euphoride: NIT: It might be advisable to deconstruct this line into semantically distinct variables (i.e. `workstation`, `gpu`, etc).

frontend/src/App.test.tsx
```diff
@@ -74,7 +82,7 @@ test("data is fetched again after refresh interval", async () => {
   // to rerender after every fetch (just in case)
   view.rerender(<App />);
 
-  const new_temp = screen.getByText("100 °C", { exact: false });
+  const new_temp = await screen.findByText("100 °C", { exact: false });
```
Euphoride: NIT: `temp` name here is not descriptive

Euphoride (approved):
LGTM besides the little nits I've put in! PR is mergeable if we're in a rush, but might be worth taking some time to dust the variable names.

frontend/src/App.test.tsx
NathanielB123: I actually meant `temp` for `temperature` not `temporary` lol, but it does slightly jump out. Renamed to `gpu_temp`.

frontend/src/App.test.tsx
NathanielB123: Done

frontend/src/App.test.tsx
NathanielB123: See above lol, renamed to `gpu_temp`

PR #91, opened 2024-02-05
Return all new fields from the DB

### LINKED ISSUE #86: Fix `Database.LatestSeen()` not returning every row in the DB
Due to a `SELECT` statement not being complete, 0's are being spat out the web api

danielg0:
None of the the new fields from "The Great Refactor" were being checked,
so I completely missed none of them were being returned :clown:

To fix this in the future, I've used reflection inside statsNear to
ensure we check every floating point member of the struct.

Closes #86


aDotInTheVoid:
LGTM: Kinda sad that we can't just use `==`, but I guess going through the DB may change some floating point bits somehow. Oh well.

PR #92, opened 2024-02-05
backend authentication functionality

### LINKED ISSUE #94: Back-end authentication

YRMYJASKA:
Closes #94


YRMYJASKA:
Closes #92

internal/femto/authentication.go
```diff
@@ -0,0 +1,56 @@
...
+	InvalidCredientalsError = errors.New("Invalid credientals for creating an authentication token")
+)
+
+const (
+	AuthCookieField = "auth_cookie"
+)
+
+type AuthToken = string
+
+type Authenticator[T any] interface {
```
aDotInTheVoid: Could you document what `T` refers to here?

internal/femto/authentication.go
```diff
@@ -0,0 +1,56 @@
...
+	RevokeToken(token AuthToken) error
+	CheckToken(token AuthToken) bool
+}
+
+func AuthWrapGet[A any, T any](auth Authenticator[A], handle GetFunc[T]) GetFunc[T] {
+	return func(r *http.Request, l *slog.Logger) (T, error) {
+		var zero T
+		cookie, err := r.Cookie(AuthCookieField)
+		if err != nil {
+			return zero, err
```
aDotInTheVoid: Could you wrap this error to document the cookie you expected? Go's [ErrNoCookie](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/net/http/request.go;l=420) doesn't have this detail

internal/femto/authentication.go
```diff
@@ -0,0 +1,56 @@
...
+type Authenticator[T any] interface {
+	CreateToken(T) (AuthToken, error)
+	RevokeToken(token AuthToken) error
+	CheckToken(token AuthToken) bool
+}
+
+func AuthWrapGet[A any, T any](auth Authenticator[A], handle GetFunc[T]) GetFunc[T] {
+	return func(r *http.Request, l *slog.Logger) (T, error) {
+		var zero T
+		cookie, err := r.Cookie(AuthCookieField)
```
aDotInTheVoid: Could we use HTTP Bearer authorization, instead of a cookie, as that's less browser specific?

internal/femto/authentication.go
```diff
@@ -0,0 +1,56 @@
...
+			return zero, NotAuthenticatedError
+		}
+		return handle(r, l)
+	}
+}
+
+func AuthWrapPost[A any, T any, R any](auth Authenticator[A], handle PostFunc[T, R]) PostFunc[T, R] {
+	return func(data T, r *http.Request, l *slog.Logger) (R, error) {
+		var zero R
+		cookie, err := r.Cookie(AuthCookieField)
```
aDotInTheVoid: NIT: This could probably be extracted out into it's own function, to deduplicate.

internal/femto/femto.go
```diff
@@ -48,6 +48,12 @@ func doGet[T any](f *Femto, w http.ResponseWriter, r *http.Request, handle GetFu
 
 	data, err := handle(r, log)
 	if err != nil {
+		// Handle authentication related errors
+		if err == NotAuthenticatedError {
```
aDotInTheVoid: This should use [errors.Is](https://pkg.go.dev/errors#Is), to support wrapping

internal/femto/femto.go
```diff
@@ -85,15 +91,36 @@ func doPost[T any](f *Femto, w http.ResponseWriter, r *http.Request, handle Post
...
+	ise := func(ctx string, e error) {
+		log.Error(ctx, "err", e)
+		http.Error(w, e.Error(), http.StatusInternalServerError)
+	}
+
+	data, userErr := handle(reqData, r, log)
+
 	if userErr != nil {
+		// Handle authentication related errors
+		if userErr == NotAuthenticatedError {
```
aDotInTheVoid: Ditto wrt `errors.Is`

internal/webapi/server.go
```diff
@@ -18,12 +18,30 @@ type api struct {
 	db database.Database
 }
 
-func NewServer(db database.Database) *Server {
+type APIAuthPacket struct {
```
aDotInTheVoid: Why's this called a Packet? I think `AuthCredentials` makes more sense. Packet is something that exists in TCP/IP, but not up here.

cmd/control/control.go
```diff
@@ -59,6 +68,38 @@ func initialiseDatabase(conf config.Database) (database.Database, error) {
 	}
 }
 
+func AuthenticatorFromConfig(config config.ControlConfiguration) configFileAuthenticator {
```
aDotInTheVoid: This is racy! If one request creates a new auth token, while another read it, you'll read from a concurrently updating map, which is bad!

Either this needs a mutex, or we should add a mutex to `femto.Femto` for the calls to `Authenticator`. I think I'd prefer the former, as we don't want to lock the mutex when doing hashing (when we do that).

Before doing this, right a test for it. See 0401d4c57b492bde063c38b71477df9e7f2a5de6 for an example of that.

YRMYJASKA: Thanks for pointing this out, kinda forgot that I was writing concurrent code!

internal/femto/authentication.go
YRMYJASKA: Im happy to do this, but I think we have to do a small rework on femto return types so I will delegate this to another PR later on. Paging @Euphoride : any opinions on Bearer?

internal/femto/authentication.go
```diff
@@ -0,0 +1,50 @@
...
+		f := AuthWrapPost(auth, func(zero struct{}, r *http.Request, l *slog.Logger) (T, error) {
+			return handle(r, l)
+		})
+		return f(struct{}{}, rr, ll)
+	}
+}
+
+func AuthWrapPost[A any, T any, R any](auth Authenticator[A], handle PostFunc[T, R]) PostFunc[T, R] {
+	return func(data T, r *http.Request, l *slog.Logger) (R, error) {
+		var zero R
```
Euphoride, line 38: This `R` type is concerning. From what I've gathered this has been added due to the authentication token providing endpoint being a POST endpoint. This really shouldn't be the case.

POST semantics are `a -> ()`. GET semantics are `a -> b`. That is to say, POST is side-effecting, GET is not, at least from the perspective of the user. This is a useful lie, as when we try to get the auth token from the server, we need to alter state in such a way that the auth token is remembered. What is important, is the state transition **is not** a function of the data provided by the user. In the case of UUID generation, it's a function of random entropy. 

For the user, this can be modelled and reasoned by a some form of pre-generated tokens, or whatever else is useful. However the user chooses to model these semantics, the user need not worry about needing to reason about state transitions given they're entirely local to the groundstation. From the perspective of the user, I would expect the API to be GET.

This prevents the heavy abstraction leakage we see here. Almost no other POST endpoints will need to return any sort of data, and just act as side-effecting endpoints. We will have an absurd amount of endpoints that need to explicitly say that they don't do any sort of pure computation, rather than this being implicit in the type. We can already see that in the changes in the groundstation post endpoints. 

We'd significantly reduce the size of this PR by setting the auth endpoint to be GET.

internal/femto/server_test.go
```diff
@@ -34,8 +35,8 @@ func TestWrongMethod(t *testing.T) {
 
 	mux := new(femto.Femto)
 
-	femto.OnPost[struct{}](mux, "/postme", func(s struct{}, r *http.Request, l *slog.Logger) error {
-		return nil
+	femto.OnPost[struct{}, struct{}](mux, "/postme", func(s struct{}, r *http.Request, l *slog.Logger) (struct{}, error) {
+		return struct{}{}, nil
```
Euphoride: We see this `return ((), ...)` here a lot due to these changes.

internal/femto/server_test.go
```diff
@@ -150,3 +151,96 @@ func TestGetApplicationErr(t *testing.T) {
...
+
+func (auth TestAuthenticator) RevokeToken(token femto.AuthToken) error {
+	return nil
+}
+
+func (auth TestAuthenticator) CheckToken(token femto.AuthToken) bool {
+	return token == "token"
+}
+
+func TestValidAuthentication(t *testing.T) {
```
Euphoride, line 169: Worth lifting the HTTP side-effect testing scaffolding from `internal/groundstation/gpustats_test.go` to test other paths, like corrupted channels.

internal/groundstation/gpustats.go
```diff
@@ -7,30 +7,34 @@ import (
...
+)
+
+func (gs *groundstation) gpustats(data uplink.GpuStatsUpload, req *http.Request, log *slog.Logger) (string, error) {
 	log.Info("Got GPU stats", "stats", data.Stats)
 
 	// NOTE: just commented this during the big refactor -jyry
 	err := gs.db.UpdateLastSeen(data.Hostname)
 	if err != nil {
-		return err
+		return "", err
```
Euphoride: `return ((), ...)` again :(

internal/groundstation/gpustats.go
```diff
@@ -7,30 +7,34 @@ import (
...
 	if len(data.Stats) > 0 {
 		err := gs.handleGPUStatSamples(data.Hostname, data.Stats)
 		if err != nil {
-			return err
+			return "", err
 		}
 	}
 
-	return nil
+	return GPUStatsOkayResponse, nil
```
Euphoride: This is in effect returning some type `T = (A, Error)` where `A = Okay | ()` which is just `Maybe Okay` and `Error = String | ()`. Issue is, this expressiveness isn't fully used. The code has the invariant that A is of type `Okay` iff Error is of type `()`, and vice versa. This is another reason this is concerning.

internal/webapi/authenticators_test.go
```diff
@@ -0,0 +1,48 @@
...
+func TestConfigFileAuthenticatorRace(t *testing.T) {
+	t.Parallel()
+
+	var wg sync.WaitGroup
+	toSpawn := 750
+	failed := false
+
+	auth := webapi.ConfigFileAuthenticator{
+		Username:      "joe",
+		Password:      "mama",
```
Euphoride, line 22: :)

internal/femto/authentication.go
```diff
@@ -0,0 +1,50 @@
...
+		cookie, err := r.Cookie(AuthCookieField)
+		if err != nil {
+			return zero, fmt.Errorf("Could not find cookie `%s`: %w", AuthCookieField, err)
+		}
+
+		res := auth.CheckToken(cookie.Value)
```
Euphoride: This is really the only non-boilerplate computation this HOF should do I think. Anything else might need rethinking.

Euphoride (changes requested):
Some thoughts (and detailed explanation to thoughts) on the structure of the PR

internal/femto/authentication.go
```diff
@@ -36,12 +36,14 @@ func AuthWrapGet[A any, T any](auth Authenticator[A], handle GetFunc[T]) GetFunc
 func AuthWrapPost[A any, T any, R any](auth Authenticator[A], handle PostFuncPure[T, R]) PostFuncPure[T, R] {
 	return func(data T, r *http.Request, l *slog.Logger) (R, error) {
 		var zero R
-		cookie, err := r.Cookie(AuthCookieField)
-		if err != nil {
-			return zero, fmt.Errorf("Could not find cookie `%s`: %w", AuthCookieField, err)
+		authorisation := r.Header.Get("Authorization")
+		split := strings.Split(authorisation, "Bearer ")
```
aDotInTheVoid: NIT: This should be strings.CutPrefix. https://pkg.go.dev/strings#CutPrefix

danielg0 (approved):
LGTM

Euphoride:
LGTM fwiw, I think we should think about restructuring the architecture for femto and femto-adjacent things

Euphoride (approved):

NathanielB123:
The coverage check is actually a menace

PR #83, opened 2024-02-01
cleanup `config` loader

aDotInTheVoid:
Don't merge untill new deployment is in.

Best reviewed commit by commit

PR #96, opened 2024-02-06
config: Remove `Merge`.

aDotInTheVoid:
We don't use it, as `toml` does this with reflection instead. It's rather unsavory code that's easy to fall out of sync with the definitions, so it's safer to not have it if it's unused.


aDotInTheVoid:
This is a part of #83, which I'm now trying to land, but in smaller PR's.

danielg0 (approved):
LGTM

PR #97, opened 2024-02-07
Frontend Authentication/Sign In

NathanielB123:
Plus some infrastructure  for supporting links to specific pages


frontend/src/App.tsx
```diff
@@ -1,103 +1,105 @@
...
 import { Navbar } from "./Components/Navbar";
 import { ColumnGrid } from "./Components/ColumnGrid";
 import { WorkStationGroup } from "./Data";
+import React, { useState } from "react";
+import { BrowserRouter, Navigate, Route, Routes } from "react-router-dom";
+import { MainView } from "./Pages/MainView";
+import { STATS_PATH } from "./Config/Paths";
 
-const API_URL = "";
+export const API_URL = "http://localhost:8000";
```
danielg0: I know this is still draft but this change will break deployment. I think we either need a way to use Docker for dev, or some way to make this configurable

NathanielB123: This is a good point. I have no idea what the best way to handle this would be though. @aDotInTheVoid @Euphoride any ideas?

Euphoride: 100% should be an environment variable. Unsure whether the current docker configuration uses this to provide the API URL but this is the easiest way that comes to mind for dev/prod.

aDotInTheVoid: yeah, it’ll depend on your build system, but their should be a pay to pull out env vars at build time

NathanielB123:
(Merge commit because Jacob has branched off this branch and I am NOT dealing with having to rebase both branches)

aDotInTheVoid:
woo! now let’s never do this ever again

PR #99, opened 2024-02-07
Deploy on HTTPS

### LINKED ISSUE #87: Deploy site using HTTPS
Using the DNS API from Hetzner and a Caddy extension.

Will require us to start storing secrets on deployment (presumably using Docker Secrets).

danielg0:
Make it so site deploys over HTTPS. Also add configuration so that satellites default to using HTTPS. We prove ownership of the domain through DNS challenges - the hetzner dns api key needs to be put in secrets/keys

Currently, we aren't allowing connections to the web API from anywhere other than the frontend. We probably want to change this eventually so our API can be used by other services. This will need some sort of per-action token to avoid being susceptible to cross-site request forgery

Closes #87


secrets/keys
```diff
@@ -0,0 +1 @@
+HETZNER_DNS_API_TOKEN=
```
aDotInTheVoid: This defiantly shouldn't be commited, if it's `.gitignore`d. If you want an give an example of how to do this, it should be named something like `secrets/keys.example`.

satellite.toml.default
```diff
@@ -1,6 +1,7 @@
 [Groundstation]
+protocol = "https"
```
aDotInTheVoid, line 2: I don't think this needs to be addressed now, but I don't like how we've got these `.default` files, which are completely seperate to the `Default()` methods in the code. I know why, but it's unintuitive, and should be changed at some point. Probably doesn't need fixing now though.

aDotInTheVoid:
r=me with secrets file renamed

satellite.toml.default
danielg0: I agree, but yeah, different pr

secrets/keys
danielg0: Sure, that's a misunderstanding on my part. I saw that, when modified, `git add secrets` and `git add secrets/keys` would fail and assumed the same applied to `git add .`

danielg0: Git moment

PR #101, opened 2024-02-08
Onboarding over SSH

aDotInTheVoid:
This implements the `onboard` package, which can SSH into a machine,
SCP the binary and config file over, and then start the satellite.

It doesn't do the `webapi` integration, which will be done in a
followup PR. For this reason, please focus your review on
`internal/onboard`, and not `cmd/onboard`, which was mainly for testing.
I know it hardcodes a config, this'll change when I do `webapi` integration.

Big outstanding questions:
- How to test this?
- How to handle the satellite already running?


YRMYJASKA (commented):
LGTM

danielg0 (commented):
> How to handle the satellite already running?

If we're not getting heartbeats, could we assume the satellite is malfunction and start our on-boarding by running `killall satellite` (this would also work for updates)? (not essential for this pr)

In terms of review, I'm glad we are putting the binary in a place other than home dir, but could it be made configurable/more obvious that it needs to be somewhere temporary that isn't on a network fs (as if we are, then multiple simultaneous onboardings could lead to one satellite running a binary that's being overwritten by an onboarding to a different satellite)

Perhaps we could use https://pkg.go.dev/os#MkdirTemp, create a temp dir and copy across the binary and config and run, after which point we don't care if the binary & config are removed because they're in memory. This does mean we'd have to onboard on every reboot, but I think that's fine (Ben's mentioned how that's a rare event, and once we have automated onboarding it becomes a good way to ensure we don't build up a bunch of old satellites on any machine)

^ This last paragraph isn't a suggestion for this pr, happy to merge without it

aDotInTheVoid:
`os.MkTemp` won't make a temp-dir on the remote machine though, which is what we need.

Your right that the `remoteDataDir` needs to be an argument though. I'll add that, then merge

PR #102, opened 2024-02-08
Backend Refactor

Euphoride:
Merged off the alerts branch, I streamlined the code for everything backend related where I could to better define dependencies and APIs. Should lead to more natural testing, and packaging.


internal/types/types.go
```diff
@@ -0,0 +1,3 @@
+package types
+
+type Unit struct{}
```
aDotInTheVoid, line 3: This feels kinda superfluous, given that `types.Unit{}` is longer than `struct{}{}`

internal/authentication/authentication.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/gpuctl/gpuctl/internal/types"
+)
+
+var (
+	NotAuthenticatedError   = errors.New("User does not have a valid authentication token")
+	InvalidCredientalsError = errors.New("Invalid credientals for creating an authentication token")
+)
+
+const (
+	AuthCookieField = "auth_cookie"
```
aDotInTheVoid: NIT: This can be removed, it's not used

internal/authentication/authentication.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type AuthToken = string
```
aDotInTheVoid, line 17: NIT: If this is in the `authentication` package, it can just be called `Token`. IE `authentication.Token` not `authentication.AuthToken`

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
 	reqNo atomic.Uint64
 }
 
-func (f *Femto) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	f.mux.ServeHTTP(w, r)
+type HTTPResponseContent[T any] struct {
+	Headers map[string]string
+	Body    T
+	Status  int
 }
```
aDotInTheVoid, line 23: Worth having `type EmptyResponce HTTPResponseContent[struct{}]`, rather than having `types.Unit` or `struct{}` everywhere

internal/database/unit_test.go
```diff
@@ -26,7 +26,7 @@ var UnitTests = [...]unitTest{
 	{"AppendingFailsIfContextMissing", appendingFailsIfContextMissing},
 	{"AppendedDataPointsAreSaved", appendedDataPointsAreSaved},
 	{"MultipleHeartbeats", multipleHeartbeats},
-	unitTest{"TestSuccessfulDrop", dropSuccess},
+	unitTest{"TestSuccessfulDrop", dropSuccess}, // Must be last unit test :)
```
aDotInTheVoid: NIT: Worth documenting why here.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
 
-func OnPost[T any, R any](f *Femto, pattern string, handle PostFuncPure[T, R]) {
-	f.mux.HandleFunc(pattern, func(w http.ResponseWriter, r *http.Request) {
-		doPost(f, w, r, handle)
+type PostFunc[T any] func(T, *http.Request, *slog.Logger) (*HTTPResponseContent[types.Unit], error)
+type GetFunc[T any] func(*http.Request, *slog.Logger) (*HTTPResponseContent[T], error)
+
+func (femto *Femto) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	femto.mux.ServeHTTP(w, r)
```
aDotInTheVoid, line 30: Single character receivers are idiomatic in Go. https://go.dev/wiki/CodeReviewComments#receiver-names. Changing this doesn't improve clairity, but does make the git diff more confusing to review.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
 )
 
 type Femto struct {
 	mux   http.ServeMux
 	reqNo atomic.Uint64
 }
 
-func (f *Femto) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	f.mux.ServeHTTP(w, r)
+type HTTPResponseContent[T any] struct {
```
aDotInTheVoid: This name is way too long given that it's used everywhere. Can we call it `femto.Response`?

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
+	femto.mux.HandleFunc(pattern, func(writer http.ResponseWriter, request *http.Request) {
+		doPost(femto, writer, request, handle)
 	})
 }
 
-func OnGet[T any](f *Femto, pattern string, handle GetFunc[T]) {
-	f.mux.HandleFunc(pattern, func(w http.ResponseWriter, r *http.Request) {
-		doGet(f, w, r, handle)
+func OnGet[T any](femto *Femto, pattern string, handle GetFunc[T]) {
+	femto.mux.HandleFunc(pattern, func(writer http.ResponseWriter, request *http.Request) {
```
aDotInTheVoid: DITTO: These renames don't improve anything, but cloud the diff.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
-	reqNo := f.nextReqNo()
-	log := f.logger().With(slog.Uint64("req_no", reqNo))
+func generateErrorLogger(log *slog.Logger, writer http.ResponseWriter) func(ctx string, status int, e error) {
+	return func(ctx string, status int, e error) {
+		log.Error(ctx, "err", e)
+		http.Error(writer, e.Error(), status)
+	}
+}
+
+func doGet[T any](femto *Femto, writer http.ResponseWriter, request *http.Request, handle GetFunc[T]) {
```
aDotInTheVoid: DITTO wrt renames. Makes it hard to see whats changed due to renaming vs others at a glance.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
-		ise("application error", err)
+		ise("There was an error in the handler", data.Status, err)
 		return
 	}
 
-	jsonb, err := json.Marshal(data)
+	jsonb, err := json.Marshal(data.Body)
 	if err != nil {
-		ise("marshaling to json failed", err)
+		ise("There was an error in trying to parse the handler's response into JSON", data.Status, err)
```
aDotInTheVoid: But we're not parsing to JSON, we're marshaling. Please keep the reference to marshaling, even if you feel the need to change the errors.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
-	_, err = w.Write(jsonb)
+	for key, value := range data.Headers {
+		writer.Header().Add(key, value)
+	}
+
+	_, err = writer.Write(jsonb)
+
 	if err != nil {
-		ise("writing failed", err)
+		ise("There was an error in trying to respond to the user", data.Status, err)
```
aDotInTheVoid: This is less usefull, as it could now be any error. This should explicity say it was a problem writing.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
 	}
 }
 
-func doPost[T any, R any](f *Femto, w http.ResponseWriter, r *http.Request, handle PostFuncPure[T, R]) {
+func doPost[T any](f *Femto, writer http.ResponseWriter, reader *http.Request, handle PostFunc[T]) {
```
aDotInTheVoid: DITTO wrt renaming churn.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
 
 	// TODO: Ensure all field's are present. I'm so mad this is hard.
 	var reqData T
-	decoder := json.NewDecoder(r.Body)
+	decoder := json.NewDecoder(reader.Body)
+
 	if err := decoder.Decode(&reqData); err != nil {
 		log.Info("Failed to unmarshal JSON", "error", err)
-		http.Error(w, "failed to decode json: "+err.Error(), http.StatusBadRequest)
+		http.Error(writer, "Failed to decode the provided JSON"+err.Error(), http.StatusBadRequest)
```
aDotInTheVoid: NIT: Their should be a `: ` here.

internal/femto/femto.go
aDotInTheVoid: Also, it's not a reader, it's a request.

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
-		if errors.Is(userErr, NotAuthenticatedError) {
-			log.Info("Invalid attempt at authentication")
-			http.Error(w, "Not authenticated", http.StatusUnauthorized)
-			return
-		}
-		// TODO: Nicer error
-		log.Info("Error", "err", userErr)
-		http.Error(w, userErr.Error(), http.StatusInternalServerError)
+
+		ise("There was an error in trying to authenticate the user", data.Status, userErr)
```
aDotInTheVoid: I don't think this message is right. The error from the handler may not be anything to do with auth

internal/femto/femto.go
```diff
@@ -3,124 +3,145 @@ package femto
...
+	for key, value := range data.Headers {
+		writer.Header().Add(key, value)
 	}
 
-	_, err = w.Write(jsonb)
+	_, err := writer.Write(make([]byte, 0))
+
 	if err != nil {
-		ise("writing failed", err)
+		ise("There was an error in trying to respond to the user", data.Status, err)
```
aDotInTheVoid: DITTO wrt usefullness of this error

internal/femto/server_test.go
```diff
@@ -195,63 +212,39 @@ func TestValidAuthentication(t *testing.T) {
 
 	mux.ServeHTTP(w, reqGet)
 	assert.Equal(t, http.StatusOK, w.Code)
-	assert.Contains(t, string(w.Body.Bytes()), "OKGET")
+	assert.Contains(t, w.Body.String(), "OKGET")
```
aDotInTheVoid, line 213: nice!

internal/groundstation/heartbeat.go
```diff
@@ -5,11 +5,17 @@ import (
...
+	"github.com/gpuctl/gpuctl/internal/types"
 	"github.com/gpuctl/gpuctl/internal/uplink"
 )
 
-func (gs *groundstation) heartbeat(data uplink.HeartbeatReq, req *http.Request, log *slog.Logger) error {
+func (gs *groundstation) heartbeat(data uplink.HeartbeatReq, req *http.Request, log *slog.Logger) (*femto.HTTPResponseContent[types.Unit], error) {
 	log.Info("Received a heartbeat", "satellite", data.Hostname)
 
-	return gs.db.UpdateLastSeen(data.Hostname, time.Now().Unix())
+	// ! Yes, I am making an exception here. The response to a heartbeat is powered by femto, but should be lightweight,
```
aDotInTheVoid: An excpetion to what! Also this references `FailHandler`, which doesn't seem to exist. Also what's going on with `// !` comments.

internal/groundstation/heartbeat.go
```diff
@@ -5,11 +5,17 @@ import (
...
+	// ! and should not rely on femto's FailHandler.
+
+	// ! Even if the heartbeat isn't recorded on our side properly, what good is it telling them that?
```
aDotInTheVoid: I'd rather ignore this on the satellite side, that do this here.

internal/webapi/alerts.go
```diff
@@ -18,5 +20,5 @@ func (wa *api) HandleOfflineMachineRequest(req *http.Request, log *slog.Logger)
 		names = append(names, machine_data[idx].Hostname)
 	}
 
-	return names, nil
+	return &femto.HTTPResponseContent[[]string]{Body: names}, nil
```
aDotInTheVoid: Doesn't this also need to set the responce code to 200? 

(The correct answer is to have `func Ok[T any](content T) Responce[T]` in `femto`)

internal/webapi/server.go
```diff
@@ -1,43 +1,49 @@
...
 
 type Server struct {
 	mux *femto.Femto
-	api *api
+	api *Api
 }
 
-type api struct {
-	db database.Database
+type Api struct {
```
aDotInTheVoid, line 19: What's the point in making all this public?

internal/webapi/server.go
```diff
@@ -1,43 +1,49 @@
...
 }
 
 type APIAuthCredientals struct {
 	Username string
 	Password string
 }
 
-func NewServer(db database.Database, auth femto.Authenticator[APIAuthCredientals]) *Server {
+// ! Key change
+const cookieFormat = "token=%s; Path=/; HttpOnly; Secure; SameSite=Strict"
```
aDotInTheVoid: This should be `func makeAuthCookie(token string) string`, so that the linters can check we have the right args to `fmt.Sprintf` (and it's easier to ensure it's used correctly)

internal/webapi/server.go
```diff
@@ -78,28 +84,36 @@ func (a *api) allstats(r *http.Request, l *slog.Logger) (workstations, error) {
 	}
 
 	result := []workstationGroup{{Name: "Shared", WorkStations: ws}}
-	return result, nil
+	return &femto.HTTPResponseContent[workstations]{Body: result}, nil
```
aDotInTheVoid: Ditto WRT respnce code

internal/webapi/server.go
```diff
@@ -49,8 +55,8 @@ func (s *Server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // This function involves a lot of weird unwrapping
 // TODO: See if we can get the database layer to do it for us
-func (a *api) allstats(r *http.Request, l *slog.Logger) (workstations, error) {
-	data, err := a.db.LatestData()
+func (a *Api) AllStatistics(r *http.Request, l *slog.Logger) (*femto.HTTPResponseContent[workstations], error) {
```
aDotInTheVoid: DITTO wrt this being public

internal/webapi/server.go
```diff
@@ -78,28 +84,36 @@ func (a *api) allstats(r *http.Request, l *slog.Logger) (workstations, error) {
...
-// TODO
-func (a *api) moveMachineGroup(move MachineMove, r *http.Request, l *slog.Logger) error {
-	l.Info("Accessed moveMachineGroup")
-	return nil
+	if err != nil {
+		return nil, err
+	}
+
+	headers := make(map[string]string)
+	headers["Set-Cookie"] = fmt.Sprintf(cookieFormat, token)
```
aDotInTheVoid: When we chech authentication, we look for a token, but we set a cookie here. Why not just return JSON with the token? That way it's also easier for non-brower clients.

internal/webapi/server.go
```diff
@@ -78,28 +84,36 @@ func (a *api) allstats(r *http.Request, l *slog.Logger) (workstations, error) {
...
-	l.Info("Accessed addMachineInfo")
-	return nil
-}
+// func (a *api) moveMachineGroup(move MachineMove, r *http.Request, l *slog.Logger) error {
+// 	l.Info("Accessed moveMachineGroup")
+// 	return nil
+// }
+
+// // TODO
+// func (a *api) addMachineInfo(info MachineAddInfo, r *http.Request, l *slog.Logger) error {
```
aDotInTheVoid, line 113: Are we doing this just to satisfy the coverage requirement? If so, I'd like to have a serious conversation about removing it, as I fear it's causing more harm than good.

internal/webapi/server.go
```diff
@@ -78,28 +84,36 @@ func (a *api) allstats(r *http.Request, l *slog.Logger) (workstations, error) {
...
+// 	l.Info("Accessed addMachineInfo")
+// 	return nil
+// }
 
 // Bodge together stats and contextual data to make OldGpuStats
-func zipStats(host string, info uplink.GPUInfo, stat uplink.GPUStatSample) OldGPUStatSample {
+func ZipStats(host string, info uplink.GPUInfo, stat uplink.GPUStatSample) OldGPUStatSample {
```
aDotInTheVoid, line 119: ditto WRT public.

aDotInTheVoid (commented):
There's alot being changed here, and it's hard to pick out the good from the bad due to it all being done in 1 commit.

It's probably a better idea to split this up into small PRs that can be reviewed 1 at a time. The only thing that requires coordingation would be changing the types of `femto.GetFunc` and `femto.PostFunc`, which should probably be done after the demo, but the rest can be done without stepping on anyone's toes.

internal/types/types.go
Euphoride: That said, `types.Unit` is way more descriptive for what it's worth. From the DX perspective, I want there to be no semantic overhead on the usage of the API. This is a tradeoff worth making

internal/authentication/authentication.go
Euphoride: I disagree. There is the potential for different tokens to be used during authentication. There exist access (authorisation) tokens, session tokens, and CSRF tokens. This is the authorisation one.

internal/femto/femto.go
Euphoride: It's not an empty response however, given headers and status are always present. Potentially this should have a name like `EmptyBodyResponse`.

internal/femto/femto.go
Euphoride: I think they're idiomatic so long as there are no more than two of them, and they are the same within the codebase. `Request` and `Response` need different character names then.

internal/femto/femto.go
Euphoride: We absolutely can. I have a bias for longer names for key API layers. We need to prioritise DX and code-clarity for layers that hide abstraction, for the same reason Go was selected.

internal/femto/femto.go
Euphoride: This is a user-facing error via `ise`. It's Go-specific error messaging to note marshalling. To the user, this is a web API and so we should prioritise idiomatic web-oriented terminology rather than that of Go.

internal/femto/femto.go
Euphoride: Responding to the user, and writing, is the same thing, no?

internal/femto/femto.go
Euphoride: True, I'll change that to noting an internal error.

internal/groundstation/heartbeat.go
Euphoride: `// !` comments show up as key comments given certain editor extensions. Also, will detail better what the exception is here.

internal/groundstation/heartbeat.go
Euphoride: I would agree with this if there was any scope on the satellite side to actually recover from a failed heartbeat. If somebody can suggest a lifecycle for that, then I'll look into propagating that error.

internal/webapi/alerts.go
Euphoride: Yes it does, which we can get put into the PR.

internal/webapi/server.go
Euphoride: It should be used in separate testing packages.

internal/webapi/server.go
Euphoride: Noted

internal/webapi/server.go
Euphoride: Testing again

internal/webapi/server.go
Euphoride: Entirely insecure. We cannot, under any circumstances, allow the client to have access to the access token. It should be a secure, HTTP-only cookie that no javascript running on the browser (or non-browser client) can use.

internal/webapi/server.go
Euphoride: The functions are unused at the moment. Would be useless writing tests for them.

internal/webapi/server.go
Euphoride: Testing :)

internal/femto/femto.go
Euphoride: > Also, it's not a reader, it's a request.

Evening brain.

internal/types/types.go
aDotInTheVoid: is is though?

i won't block merging on this, as long as it gets wrapped up inside `femto.EmptyResponce` so it's not used everywhere. but it's not descriptive at all (relies on knowing type-theory stuff), and adding an entire package so that we can have a unit type **with a longer name** than the completely descriptive `struct{}` is super terable

internal/authentication/authentication.go
aDotInTheVoid: Then lets call it `BearerToken`, because that's how it's used. Some of the tokens being used in authorization not being authorization tokens is just going to make a mess.

internal/femto/femto.go
aDotInTheVoid: I'd say the oposite. Key API layers should have short, sweet names. Femto is all about HTTP, so it's redundant to have `HTTP` in the name. Similarly, we don't have a seperate ResponceMeta or ResponceBody to contend with, so why specify Content

(aside: Also if you wanted to return HTTP 204 No contend, but set some headers, you'd still use this, so not only is it too long, it's wrong)

I'm not sure if your saying your willing to do the rename here on not. But this current name is unacceptably long, and I'm not willing to merge it. See <https://journal.stuffwithstuff.com/2016/06/16/long-names-are-long/> for more.

internal/femto/femto.go
aDotInTheVoid: Sure, but we're still not parsing. Parsing is string->structured, and this is structured->string. If you want a less go specific term, that's fine, as long as it's right.

internal/femto/femto.go
aDotInTheVoid: In my mind, "responding to the user" could be anywhere in the entire handler. I'd like the logs to note that it's an IO error around writing the responce (and not for example calculating it)

internal/groundstation/heartbeat.go
aDotInTheVoid: Just becasue we can't recover from an error, and will just ignore it, doesn't mean we should hide the error.

Let's say someone sets up an L7 proxy that sniffs HTTP status codes to monitor heath of application (a-la linkerd). We'd want it to know something was wrong.

internal/types/types.go
Euphoride: I think it's not a bad bet to assume that the programmer using this API (us) or any maintainer in the future would be familiar with the idea of unit.

internal/femto/femto.go
Euphoride: I think you've misread semantics there. Here Content = (Status, Header, Body), not Response = (Status, Header, Content). This is a case of isomoprhism up to naming, so we're really just deciding on ontology here. I wouldn't go so far as to claim its semantically "wrong". A response's content that has HTTP 204 is (204, ..., ()). 

If you're extremely keen on having your version of the naming, I'll get that done.

internal/authentication/authentication.go
Euphoride: Not quite. Other tokens, like CSRF and session tokens **are** authorisation tokens. It may be worth renaming this one, but all tokens (which are either cookies, bearers or tagged along with all responses) are authorisation tokens given they are tokens which actually necessarily do authorise they user's action.

internal/femto/femto.go
Euphoride: We'll call it serialisation then I think

internal/femto/femto.go
Euphoride: Sure, should we say an error in _communicating_ to the user then?

internal/femto/femto.go
aDotInTheVoid: Yes, please rename it. It's confusing, and too long to be so fundamental

internal/femto/femto.go
aDotInTheVoid: why not just writing?

internal/femto/femto.go
aDotInTheVoid: Yeah, it deffinatly needs a name. I don't paticularlty care, as long as it's short

internal/femto/femto.go
Euphoride: Writing is server-specific (we write to a socket), not client-specific (the server is communicating with us) and this is a client-facing API.

internal/groundstation/heartbeat.go
Euphoride: This is a choice on either fail noisy or fail silent semantics - we're certain it'd be useful to fail-noisy?

internal/femto/femto.go
aDotInTheVoid: But if we're failing to write, the client isn't going to see it. These errors also go to our logs, and they need to be usefull their aswell.

(but whatever, we've burnt too many cycles on this, let's get the important things right)

internal/groundstation/heartbeat.go
aDotInTheVoid: Not certain, but I don't see any compelling reason to hide this error.

internal/femto/server_test.go
```diff
@@ -119,14 +119,15 @@ func TestGetHappyPath(t *testing.T) {
 	assert.JSONEq(t, `{"X": 101}`, string(j))
 }
 
-type unit struct{}
+type unit = struct{}
```
aDotInTheVoid: What's this doing?

internal/femto/server_test.go
```diff
@@ -119,14 +119,15 @@ func TestGetHappyPath(t *testing.T) {
...
+type other_unit = struct{}
```
aDotInTheVoid: NIT: Go nameing says this should be `otherUnit`, but also it should just be deleted

internal/femto/femto.go
```diff
@@ -21,9 +21,9 @@ type Response[T any] struct {
 	Status  int
 }
 
-type EmptyBodyResponse Response[types.Unit]
+type EmptyBodyResponse = Response[types.Unit]
```
aDotInTheVoid, line 24: This name is too long to be so common.

internal/femto/server_test.go
```diff
@@ -119,14 +119,15 @@ func TestGetHappyPath(t *testing.T) {
...
 
 func TestGetWrongMethod(t *testing.T) {
 	t.Parallel()
 
 	mux := new(femto.Femto)
-	femto.OnGet(mux, "/api", func(r *http.Request, l *slog.Logger) (*femto.Response[types.Unit], error) {
-		return &femto.Response[types.Unit]{}, nil
+	femto.OnGet[types.Unit](mux, "/api", func(r *http.Request, l *slog.Logger) (*femto.EmptyBodyResponse, error) {
```
aDotInTheVoid: Does this not infer?

internal/webapi/server_test.go
```diff
@@ -1 +1,49 @@
...
+
+// 	"github.com/gpuctl/gpuctl/internal/authentication"
+// 	"github.com/gpuctl/gpuctl/internal/database"
+// 	"github.com/gpuctl/gpuctl/internal/webapi"
+// )
+
+// func TestAuthenticate(t *testing.T) {
+// 	// Create a mock logger for testing
+// 	mockLogger := slog.Default()
+
```
aDotInTheVoid: What's this doing here?

internal/webapi/server_test.go
```diff
@@ -1 +1,49 @@
 package webapi_test
+
+// import (
+// 	"log/slog"
+// 	"net/http"
+// 	"net/http/httptest"
+// 	"testing"
```
YRMYJASKA: I think this file should be removed as it is useless

YRMYJASKA (commented):
LGTM. I think `Response[T]` is quite needed if we want to move away from the current authentication pipeline.

PR #104, opened 2024-02-08
Set more CORS headers in dev-mode

aDotInTheVoid:
This should be reviewed here, but may land on nathaniels auth branch.

### DIFF
--- a/Caddyfile
+++ b/Caddyfile
@@ -11,9 +11,16 @@ https://gpuctl.perial.co.uk {
 
 		X-Clacks-Overhead "GNU Terry Pratchett"
 
-		# TODO: allow the API to be accessed from other endpoints
-		# This will need CSRF protection
+		# When serving under the webapi, we only want API requests
+		# from our own origin. So remove the CORS headers that are
+		# to allow cross-origin for local-dev
+
+		# If we want to allow cross-origin api requests, we need
+		# to think carefully about CSRF, but we don't need that for
+		# non-browser stuff, so it's fine.
 		-Access-Control-Allow-Origin
+		-Access-Control-Allow-Methods
+		-Access-Control-Request-Headers
 	}
 
 	# For api routes, we can lock down the CSP even further
--- a/Caddyfile
+++ b/Caddyfile
@@ -20,7 +20,7 @@ https://gpuctl.perial.co.uk {
 		# non-browser stuff, so it's fine.
 		-Access-Control-Allow-Origin
 		-Access-Control-Allow-Methods
-		-Access-Control-Request-Headers
+		-Access-Control-Allow-Headers
 	}
 
 	# For api routes, we can lock down the CSP even further
--- a/internal/femto/femto.go
+++ b/internal/femto/femto.go
@@ -48,6 +48,7 @@ func correctMethod(method string, req *http.Request, w http.ResponseWriter, log 
 
 		w.Header().Set("Allow", method)
 		w.Header().Set("Access-Control-Allow-Methods", method)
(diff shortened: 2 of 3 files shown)


aDotInTheVoid:
Marking as draft so it doens't get merged.

internal/femto/femto.go
```diff
@@ -30,15 +30,42 @@ func OnGet[T any](f *Femto, pattern string, handle GetFunc[T]) {
...
+
+		// We don't set Access-Control-Allow-Origin, as that'd done globally
+		// so it can be enabled for dev, but not prod.
+
+		// https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request
+		// https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/OPTIONS
+
+		w.Header().Set("Allow", method)
+		w.Header().Set("Access-Control-Allow-Methods", method)
+		w.Header().Set("Access-Control-Allow-Headers", "Authorization, Content-Type")
```
danielg0, line 51: `Content-Type` is a [CORS-safelisted Request Header](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header) that's always allowed, so doesn't have to be listed in `Access-Control-Allow-Headers` unless we want to bypass the additional restrictions (ie. we want to allow non-mime types here)

Caddyfile
```diff
@@ -11,9 +11,16 @@ https://gpuctl.perial.co.uk {
...
+		# When serving under the webapi, we only want API requests
+		# from our own origin. So remove the CORS headers that are
+		# to allow cross-origin for local-dev
+
+		# If we want to allow cross-origin api requests, we need
+		# to think carefully about CSRF, but we don't need that for
+		# non-browser stuff, so it's fine.
 		-Access-Control-Allow-Origin
+		-Access-Control-Allow-Methods
+		-Access-Control-Allow-Headers
```
danielg0, line 23: I don't think we want to remove these headers if we want this to work in production. I remove the `Access-Control-Allow-Origin` header because in dev, we allow for requests to be made from `localhost:3000`, which never happens in prod. However, if I've read MDN right, `-Methods` and `-Headers` apply to all requests, not just ones from different origins

internal/femto/femto.go
aDotInTheVoid: Hmmm, for some reason this mattered though:

image

image

And the `Authorization` header wasn't set in the request we were fighting with. I just added it because I figured it'd come up.

Caddyfile
aDotInTheVoid: Oooh, where does it say that?

internal/femto/femto.go
danielg0: Did you try adding the header, but making it empty?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers seems to suggest it's required to be at least present when a [`Access-Control-Request-Headers`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Headers) is in a preflight request

Caddyfile
danielg0: It doesn't explicitly say that, but compared to the page for https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin, the pages for https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Request-Headers and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods make no reference to origin at all, implying they apply to all requests

aDotInTheVoid:
This has landed via other means

PR #111, opened 2024-02-10
ci: Ensure that `go test -short` works without postgres.

aDotInTheVoid:
Currently, tests that relly on having a DB are skipped when `t.Short()`
is true. However, this isn't enforced by CI at all. By adding this
check, we ensure this will always be the case.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -86,4 +86,13 @@ jobs:
       - name: Check Frontend Formatting
         run: npm run fmt-check
         working-directory: ./frontend
-      
+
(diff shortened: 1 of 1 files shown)


danielg0 (approved):
LGTM

Does making a new build step have them run in parallel? If so, might it be worth splitting the npm build stuff into it's own step?

aDotInTheVoid:
Yeah, can do in a followup, if you think the current build-times are too long. But the `build-and-test` is only 1m, so it wasn't a worry.

The `docker build` will defiantly be it's own job, as it'll be long.

PR #114, opened 2024-02-10
re-enable `assets` tests.

aDotInTheVoid:
These were written to juice coverage numbers, but actually did catch a bug in assets, so I'd like to keep them.

If `CGO_ENABLED=0` isn't set when building `satellite-amd64-linux`, it'll depend on libc.

```
--- FAIL: TestSatelliteAmd64LinuxIsElf (0.00s)
    assets_test.go:22:
                Error Trace:    /home/alona/dev/gpuctl/internal/assets/assets_test.go:22
                Error:          Should be empty, but was [libc.so.6]
                Test:           TestSatelliteAmd64LinuxIsElf
                Messages:       satellite binary has imports, but should be static binary
FAIL
FAIL    github.com/gpuctl/gpuctl/internal/assets        0.002s
FAIL
```

For maximum likelihood of working on the remote machine, we'd like to ensure we only use the linux syscall interface instead.

### DIFF
--- a/internal/assets/assets_test.go
+++ b/internal/assets/assets_test.go
@@ -1,28 +1,28 @@
 package assets_test
 
-// import (
-// 	"bytes"
-// 	"debug/elf"
-// 	"testing"
+import (
+	"bytes"
+	"debug/elf"
+	"testing"
 
-// 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


Euphoride (approved):
My bad, it's removal shouldn't have happened. Might have been an artifact of some other changes. LGTM

PR #103, opened 2024-02-08
Move deployment code to it's own directory

aDotInTheVoid:

PR #115, opened 2024-02-10
Combine READMEs

danielg0:
For context, see comments on #103

TDLR: I'd prefer we keep all the various documenting on deployment, etc. to one single README.md. That way, it's easier for end-users to find and for us to keep updated


aDotInTheVoid:
I like having different documents for different things. End users don't need to see how to run the `npm` tests, or intracisies of building freestanding docker (at least not immediately).

That said, we should document how to get things running. I've added some docs in a2e22533c0820251ccd24d8e23a27150d59f9249

PR #117, opened 2024-02-10
`database`: Deterministic process list merging.

### LINKED ISSUE #113: database.TestCalculateAverage is flaky
It's failed CI here: https://github.com/gpuctl/gpuctl/actions/runs/7856261718/job/21438860855.

And can be reproduced locally:

```
alona@Ashtabula:~/dev/gpuctl$ go test -short ./internal/database/ -count 10
--- FAIL: TestCalculateAverage (0.00s)
    prune_test.go:78: calculateAverage() = {gpu-1 55 30 4500 55 72.5 67.5 1.15 110 1525 1825 725 925 1625140800 [{2 Process2 300} {1 Process1 250}]}, want {gpu-1 55 30 4500 55 72.5 67.5 1.15 110 1525 1825 725 925 1625140800 [{1 Process1 250} {2 Process2 300}]}
FAIL
FAIL    github.com/gpuctl/gpuctl/internal/database      0.017s
FAIL
```

Note: You may need to repeat this command a few times to get it to work, as it's non-deterministic. The `-count` flag is needed to run multiple times, and bypass Go's builtin cacheing

---

I've not investigated the reason, but it smells like rellying on map iteration order.

CC @Euphoride

aDotInTheVoid:
Fixes #113.

Go's `map` type randomizes iteration order, to avoid user code relying on a
particular ordering. If you want a guaranteed order, you need to sort the output.
From https://go.dev/blog/rebuild:

> Randomness. Map iteration and running work in multiple goroutines serialized
> with locks both introduce randomness in the order that results may be generated.
> This randomness can make the toolchain produce one of several different possible
> outputs each time it runs. To make the build reproducible, we had to find each
> of these and sort the relevant list of items before using it to generate output.

For now, we sort by PID.

As a followup, we should consider how to merge the same PID/name over time.

### DIFF
--- a/internal/database/inmemory.go
+++ b/internal/database/inmemory.go
@@ -1,8 +1,10 @@
 package database
 
 import (
+	"cmp"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


danielg0 (approved):
LGTM

Issue #120, opened 2024-02-12
Convert shortcode->full name, and show in frontend for process list.


aDotInTheVoid:
Or etsy???

What????

aDotInTheVoid:
`nvidia-smi` shows unix username for all running processes. We'd like to show a users full-name instead.

PR #137, opened 2024-02-12
passwd: New package to parse `/etc/passwd`

aDotInTheVoid:
First part of #120


YRMYJASKA (commented):
LGTM, no major comments. Small nit is that we use `_testdata` in our `gpustats` tests and here we use `testdata`, would be nice to be consistent.

aDotInTheVoid:
`testdata` is idiomatic, I'll submit a PR.

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests. 

-- <https://pkg.go.dev/cmd/go#hdr-Test_packages>

PR #138, opened 2024-02-12
added development script to launch the whole thing

### LINKED ISSUE #127: Create dev env scripts for with and without gpu
Currently starting the dev env involves running the satellite, control, and frontend - this could all be done by a script. We would need a separate script for when using real or fake gpu data.

YRMYJASKA:
Will add a new script `launch-dev` in the root that runs the satellite, groundstation, and front-end at the same time. Will also clean-up (i.e. kill the child processes) after it receives a `SIGINT`

Closes #127

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ satellite.toml
 .DS_Store
 cover.out
 __debug*
+*.log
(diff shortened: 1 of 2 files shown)


YRMYJASKA:
Closes #127

YRMYJASKA:
This is an isolated change, will just merge

PR #139, opened 2024-02-13
Make satellite resend context when it suspects server cannot remember gpu

### LINKED ISSUE #134: Satellite resend context
If a satellite attempts to send data but gets an error because the database is unaware of it, it should resend the context

YRMYJASKA:
Closes #134

### DIFF
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -2,10 +2,10 @@ package main
 
 import (
 	"encoding/json"
-	"log/slog"
-	"os"
 	"errors"
+	"log/slog"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/gpuctl/gpuctl/internal/config"
@@ -191,9 +191,11 @@ func (s *satellite) sendHeartBeat() error {
 		s.gsAddr+uplink.HeartbeatUrl,
 		uplink.HeartbeatReq{Hostname: s.hostname},
 	)
+
 	if err != nil {
 		return err
 	}
+
 	if resp.StatusCode != http.StatusOK {
 		return err // TODO: add errors for each status code
 	}
@@ -236,6 +238,11 @@ func (s *satellite) sendGPUStatus(stats []uplink.GPUStatSample) error {
 		s.gsAddr+uplink.GPUStatsUrl,
 		uplink.GpuStatsUpload{Hostname: s.hostname, Stats: stats},
 	)
+
+	if err != nil {
+		return err
+	}
+
 	if resp.StatusCode == http.StatusBadRequest {
 		// Could not send status, suspected cause is server does not have context.
 		return errSuspectedServerMissingInfo
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -4,12 +4,18 @@ import (
 	"encoding/json"
 	"log/slog"
 	"os"
+	"errors"
+	"net/http"
 	"time"
 
 	"github.com/gpuctl/gpuctl/internal/config"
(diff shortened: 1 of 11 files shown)


YRMYJASKA:
Closes #134

internal/database/postgres.go
```diff
@@ -176,6 +176,11 @@ func (conn postgresConn) AppendDataPoint(sample uplink.GPUStatSample) error {
 		sample.GraphicsClock, sample.MaxGraphicsClock,
 		sample.MemoryClock, sample.MaxMemoryClock)
 
+	// TODO: hacky, untested. Might not work
```
danielg0, line 179: This is fine, because I'm about to change a fair bit of postgres.go anyway

danielg0 (approved):
LGTM

This does treat any 400 response as a missing GPU error, but that's fine because that's the only time we currently return a 400. If we wanted to return 400 codes for other reasons (invalid json, etc.), it would be neat to see if we could encode the Go error in the response body, then have one set of errors shared between the groundstation and satellites (so, we can use `errors.Is()`)

internal/femto/femto.go
```diff
@@ -75,10 +75,10 @@ func Ok[T any](content T) (*Response[T], error) {
 	return &Response[T]{Status: http.StatusOK, Body: content}, nil
 }
 
-func generateErrorLogger(l *slog.Logger, w http.ResponseWriter) func(ctx string, status int, e error) {
+func generateErrorLogger(l *slog.Logger, w *http.ResponseWriter) func(ctx string, status int, e error) {
```
aDotInTheVoid, line 78: NIT: Whats the pupose of this change? [ResponceWriter](https://pkg.go.dev/net/http#ResponseWriter) is already an interface, so should be passed by value, unless you have a specific need

internal/database/inmemory.go
```diff
@@ -42,7 +37,7 @@ func (m *inMemory) AppendDataPoint(sample uplink.GPUStatSample) error {
 	defer m.mu.Unlock()
 
 	if info, pres := m.infos[sample.Uuid]; !pres {
-		return fmt.Errorf("%w: %s", ErrGpuNotPresent, sample.Uuid)
+		return ErrGpuNotPresent
```
aDotInTheVoid, line 40: Why make this change?

internal/groundstation/gpustats.go
```diff
@@ -27,7 +28,9 @@ func (gs *groundstation) gpustats(data uplink.GpuStatsUpload, req *http.Request,
 
 	if len(data.Stats) > 0 {
 		err := gs.handleGPUStatSamples(data.Hostname, data.Stats)
-		if err != nil {
+		if err == database.ErrGpuNotPresent {
```
aDotInTheVoid, line 31: This should use `errors.Is`, so it works on wrapped errors.

cmd/satellite/main.go
```diff
@@ -194,14 +229,24 @@ func (s *satellite) sendGPUStatusWithSource(gpuhandler gpustats.GPUDataSource) e
...
 		uplink.GpuStatsUpload{Hostname: s.hostname, Stats: stats},
 	)
+
+	if err != nil {
+		return err
+	}
+
+	if resp.StatusCode == http.StatusBadRequest {
+		// Could not send status, suspected cause is server does not have context.
+		return errSuspectedServerMissingInfo
```
aDotInTheVoid, line 248: NIT: Needs a todo that this assumes only 1 error.

aDotInTheVoid (commented):
Gah!, I thaught I'd reviewed this, but github apparenly kept my comments in the pending state?

PR #142, opened 2024-02-13
database: Remove `Drop` from database interface.

aDotInTheVoid:
Why? We don't want this called from application code, only tests. To enforce this, we make
it a method on `PostgresConn`, instead of `Database`.

This requires a number of changes:

1. Make `PostgresConn` public
2. Make `Postgres` return a `*PostgresConn`, so it tests can see methods not on the interface.
  - We could have it return a `PostgresConn`, but then it isn't nil-able, so error handling suffers.
3. Make `*PostgresConn` implement `Database` (instead of `postgresConn`), so we can use pointer return.

### DIFF
--- a/internal/database/inmemory.go
+++ b/internal/database/inmemory.go
@@ -6,7 +6,6 @@ import (
 	"slices"
 	"sort"
 	"sync"
-	"testing"
 	"time"
 
 	"github.com/gpuctl/gpuctl/internal/broadcast"
@@ -217,11 +216,6 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
 	}
 
 	return averagedSample
-}
-
-func (m *inMemory) Drop(t *testing.T) {
-	m.mu.Lock()
-	m = nil
 }
 
 func (m *inMemory) NewMachine(machine broadcast.NewMachine) error {
--- a/internal/database/interface.go
+++ b/internal/database/interface.go
@@ -2,7 +2,6 @@ package database
 
 import (
 	"errors"
-	"testing"
 
 	"github.com/gpuctl/gpuctl/internal/broadcast"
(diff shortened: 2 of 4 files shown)


danielg0 (commented):
Fine with the general idea, but:

- I think I'd still rather have the postgres func return a database. If I've understood this correctly, we can cast the returned database to `PostgresConn` in the tests - it's an extra bit of legwork, but I think it highlights using  Drop isn't really suggested
- if we're going to rework postgresconn, I think we should go back and just make it the typealias to `*sql.DB` it always should have been. That way, you get the nice nullability without needing two levels of indirection

aDotInTheVoid:
> I think I'd still rather have the postgres func return a database. If I've understood this correctly, we can cast the returned database to PostgresConn in the tests - it's an extra bit of legwork, but I think it highlights using Drop isn't realy suggested

That's a good idea, I'll do that

aDotInTheVoid:
> if we're going to rework postgresconn, I think we should go back and just make it the typealias to *sql.DB it always should have been. That way, you get the nice nullability without needing two levels of indirection

I disagree. Theirs no point exposing that to the users, incase we want to change them (E.G. adding a redis connection, or random metadata)

What's the advantage of being a alias/newtype?

I'm not paticularly concerned about double-indirection, given that this sort of application is all io-bound, and we only have 1 (not n) `database.Postgres`'s anyway.

danielg0:
> I'm not paticularly concerned about double-indirection, given that this sort of application is all io-bound, and we only have 1 (not n) `database.Postgres`'s anyway.

That's fair enough

aDotInTheVoid:
Now that we're returning an interface again, we can go back to using value-receivers.

danielg0 (approved):
LGTM

Issue #129, opened 2024-02-12
UUID not a string pls


aDotInTheVoid:

danielg0:
I would like to propose this issue is closed.

The UUID can be converted to a 128-bit numeric value, but neither Go nor Postgres have a type for 128-bit ints, so we would have to use arbitrary precision integers. In [Postgres](https://www.postgresql.org/docs/16/datatype-numeric.html#DATATYPE-NUMERIC-DECIMAL), these are stored as BCDs and based on some quick calculations, I don't believe changing to them would give us much of a space/speed advantage.

It's true that it guarantees that the UUID is the 128-bit number we expect, but as we don't perform any numeric operations on the value, only equality checks, I don't see why a string rep isn't appropriate.

TLDR: I don't currently see a benefit to storing UUID as not a string that outweighs the complexity cost of parsing to/from the string representation

aDotInTheVoid:
@NathanielB123 this is your issue, What Do You Think

CC @sirjacobofglais , i think this came up a lot for you last summer

sirjacobofglais:
it's (extremely) badly abstracted as a string - using that datatype is just misunderstanding what a UUID is.

if representing it any other way would be significantly clunkly then I think it's fine to keep it like this, BUT it definitely needs a long comment on the typedef to explain why we are doing this. And we should keep this issue in mind if we run into stringy problems with this in the future.

NathanielB123:
I guess the abstraction concern largely comes down to how numbers can be represented multiple ways by a string (decimal, hex, base64 etc...)? - if so, commenting/documentation sounds like a reasonable solution to me.

danielg0:
> The UUID can be converted to a 128-bit numeric value, but neither Go nor Postgres have a type for 128-bit ints

My argument here is invalidated. Whilst it is true that Postgres doesn't have a type for 128-bit integers, it has a type for UUIDs (https://www.postgresql.org/docs/16/datatype-uuid.html) :facepalm:

PR #143, opened 2024-02-14
Return groups from the database

### LINKED ISSUE #21: Attach files and notes to machines in the admin panel
When setting up/editing a machine in the admin panel, allow for admins to make notes (eg. noisy GPU fan) or upload documents (eg. invoices) that are attached to that particular machine.

Should probably be stored locally (don't want to transmit them elsewhere)

### LINKED ISSUE #76: Support Workstation Groups through the Back-End

### LINKED ISSUE #124: Fix groups

danielg0:
In order to return groups from the database, make a major change to the database interface. For "output" functions, rather than returning the uplink types (those for satellite-to-groundstation communication), return the broadcast types (those for webapi-to-frontend communication). This means we can remove all the zipping work from the front-end.

Closes #124
Closes #76
Makes progress toward closing #21 (we still want to be able to attach files)

### DIFF
--- a/internal/database/inmemory.go
+++ b/internal/database/inmemory.go
@@ -3,8 +3,10 @@ package database
 import (
 	"cmp"
 	"errors"
+	"reflect"
 	"slices"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -18,14 +20,16 @@ type gpuInfo struct {
 }
 
 type inMemory struct {
-	infos    map[string]gpuInfo                // maps from uuids to context info
-	stats    map[string][]uplink.GPUStatSample // maps from uuids to slices of stats, allowing tracking of multiple datapoints
-	lastSeen map[string]int64                  // map from hostname to last seen time
-	mu       sync.Mutex                        // mutex
+	machines map[string]broadcast.ModifyMachine // maps from hostname to machine info
+	infos    map[string]gpuInfo                 // maps from uuids to context info
+	stats    map[string][]uplink.GPUStatSample  // maps from uuids to slices of stats, allowing tracking of multiple datapoints
+	lastSeen map[string]int64                   // map from hostname to last seen time
+	mu       sync.Mutex                         // mutex
 }
 
 func InMemory() Database {
 	return &inMemory{
+		machines: make(map[string]broadcast.ModifyMachine),
 		infos:    make(map[string]gpuInfo),
 		stats:    make(map[string][]uplink.GPUStatSample),
 		lastSeen: make(map[string]int64),
@@ -61,37 +65,68 @@ func (m *inMemory) UpdateGPUContext(host string, packet uplink.GPUInfo) error {
 	return nil
 }
 
-func (m *inMemory) LatestData() ([]uplink.GpuStatsUpload, error) {
+func (m *inMemory) LatestData() (broadcast.Workstations, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	var uploads []uplink.GpuStatsUpload
-	grouped := make(map[string]*uplink.GpuStatsUpload)
-
-	for uuid, samples := range m.stats {
-		if len(samples) == 0 {
-			continue
+	// make mapping from machine->gpu, then make group heirarchy
+	var gpus = make(map[string][]broadcast.GPU)
+	for uuid, info := range m.infos {
+		gpu := broadcast.GPU{
+			Uuid:          uuid,
+			Name:          info.context.Name,
+			Brand:         info.context.Brand,
+			DriverVersion: info.context.DriverVersion,
+			MemoryTotal:   info.context.MemoryTotal,
 		}
-		info := m.infos[uuid]
 
-		if _, exists := grouped[info.host]; !exists {
-			grouped[info.host] = &uplink.GpuStatsUpload{
-				Hostname: info.host,
-				GPUInfos: []uplink.GPUInfo{info.context},
-				Stats:    []uplink.GPUStatSample{samples[len(samples)-1]}, // Latest sample
+		// most recent stat is at the end
+		stats := m.stats[uuid]
+		stat := stats[len(stats)-1]
+		// reflect on stat to get all the fields
+		for _, field := range reflect.VisibleFields(reflect.TypeOf(stat)) {
+			// uuid already set, skip
+			if field.Name == "Uuid" {
+				continue
 			}
-		} else {
-			upload := grouped[info.host]
-			upload.GPUInfos = append(upload.GPUInfos, info.context)
-			upload.Stats = append(upload.Stats, samples[len(samples)-1])
+
+			// we need a reference to gpu, not a copy so it's settable
+			target := reflect.ValueOf(&gpu).Elem().FieldByName(field.Name)
+			if target.CanSet() {
+				target.Set(reflect.ValueOf(stat).FieldByIndex(field.Index))
+			}
 		}
+
+		gpus[info.host] = append(gpus[info.host], gpu)
 	}
 
-	for _, upload := range grouped {
-		uploads = append(uploads, *upload)
+	var groups = make(map[string][]broadcast.Workstation)
+
+	for machine, info := range m.machines {
+		group := info.Group
+		if group == nil || strings.TrimSpace(*group) == "" {
+			group = &defaultGroup
+		}
+
+		var workstation = broadcast.Workstation{
+			Name:        machine,
+			CPU:         info.CPU,
+			Motherboard: info.Motherboard,
+			Notes:       info.Notes,
+			LastSeen:    time.Since(time.Unix(m.lastSeen[machine], 0)),
+			Gpus:        gpus[machine],
+		}
+
+		groups[*group] = append(groups[*group], workstation)
 	}
 
-	return uploads, nil
+	// flatten map
+	var result broadcast.Workstations = nil
+	for group, machines := range groups {
+		result = append(result, broadcast.Group{Name: group, Workstations: machines})
+	}
+
+	return result, nil
 }
 
 func (m *inMemory) UpdateLastSeen(host string, time int64) error {
@@ -99,16 +134,23 @@ func (m *inMemory) UpdateLastSeen(host string, time int64) error {
 	defer m.mu.Unlock()
 
 	m.lastSeen[host] = time
+
+	// make sure it's present in the machine list
+	_, found := m.machines[host]
+	if !found {
+		m.machines[host] = broadcast.ModifyMachine{Hostname: host}
+	}
+
 	return nil
 }
 
-func (m *inMemory) LastSeen() ([]uplink.WorkstationSeen, error) {
+func (m *inMemory) LastSeen() ([]broadcast.WorkstationSeen, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
-	var seen []uplink.WorkstationSeen
-	for name, time := range m.lastSeen {
-		seen = append(seen, uplink.WorkstationSeen{Hostname: name, LastSeen: time})
+	var seen []broadcast.WorkstationSeen
+	for name, lastSeen := range m.lastSeen {
+		seen = append(seen, broadcast.WorkstationSeen{Hostname: name, LastSeen: lastSeen})
 	}
 
 	return seen, nil
--- a/internal/database/inmemory.go
+++ b/internal/database/inmemory.go
@@ -105,7 +105,8 @@ func (m *inMemory) LatestData() (broadcast.Workstations, error) {
 	for machine, info := range m.machines {
 		group := info.Group
 		if group == nil || strings.TrimSpace(*group) == "" {
-			group = &defaultGroup
+			fallback := defaultGroup
+			group = &fallback
 		}
 
 		var workstation = broadcast.Workstation{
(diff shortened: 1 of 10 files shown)


danielg0:
Please do review if you find something wrong, but I've left it as draft for the moment because I want to:

- [x] Double check all the tests are working
- [x] Write some new tests (covering group changes propagating to the db, etc.)
- [x] Replace some of the very long comparisons with reflection
- [x] Get confirmation from @NathanielB123 that this will work right with the json the frontend's expecting

internal/database/interface.go
```diff
@@ -13,6 +13,9 @@ var (
 	ErrGpuNotPresent     = errors.New("appending to non present gpu")
 )
 
+// default group to give to machines with a null or empty group
+var defaultGroup = "Shared"
```
aDotInTheVoid: NIT: Can this be `const` instead?

danielg0: Change made in eb0287e

NathanielB123:
> Please do review if you find something wrong, but I've left it as draft for the moment because I want to:
> 
> * [x]  Double check all the tests are working
> * [x]  Write some new tests (covering group changes propagating to the db, etc.)
> * [x]  Replace some of the very long comparisons with reflection
> * [x]  Get confirmation from @NathanielB123 that this will work right with the json the frontend's expecting

On last point - only differences I can spot are:
- front-end not using `uuid` (but extra fields are handled gracefully by default iirc so shouldn't be a problem) 
- on the front-end the `workstations` field is named (extremely inconsistently) `workStations`  - not sure if JSON marshalling is case sensitive, but honestly I'd be in support of renaming the field on the front-end with this PR.

danielg0:
> * on the front-end the `workstations` field is named (extremely inconsistently) `workStations`  - not sure if JSON marshalling is case sensitive, but honestly I'd be in support of renaming the field on the front-end with this PR.

Done

danielg0:
Acc, there's a conflict, I'm going to do that in a new PR

danielg0:
> * on the front-end the `workstations` field is named (extremely inconsistently) `workStations`  - not sure if JSON marshalling is case sensitive, but honestly I'd be in support of renaming the field on the front-end with this PR.

See #148

internal/database/inmemory.go
```diff
@@ -18,14 +20,16 @@ type gpuInfo struct {
 }
 
 type inMemory struct {
-	infos    map[string]gpuInfo                // maps from uuids to context info
-	stats    map[string][]uplink.GPUStatSample // maps from uuids to slices of stats, allowing tracking of multiple datapoints
-	lastSeen map[string]int64                  // map from hostname to last seen time
-	mu       sync.Mutex                        // mutex
+	machines map[string]broadcast.ModifyMachine // maps from hostname to machine info
```
Euphoride, line 23: Is this necessary? These side-effects won't be exposed, and inMemory is only intended to be used as a mock database. Control does not read what it writes here and thus I don't believe this is necessary.

internal/database/postgres.go
```diff
@@ -60,7 +62,7 @@ func createTables(db *sql.DB) error {
 	}
 
 	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS GPUs (
-		Uuid CHAR(42) NOT NULL,
+		Uuid text NOT NULL,
```
Euphoride, line 66: Postgres should be using an extension to represent a UUID type here. We should also be using a Go UUID type, and the frontend should use a typescript UUID type.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
-	rows, err := conn.db.Query(`SELECT g.Machine, g.Uuid, g.Uuid, g.Name,
-			g.Brand, g.DriverVersion, g.MemoryTotal,
-			s.MemoryUtilisation, s.GpuUtilisation, s.MemoryUsed,
-			s.FanSpeed, s.Temp, s.MemoryTemp, s.GraphicsVoltage,
-			s.PowerDraw, s.GraphicsClock, s.MaxGraphicsClock,
-			s.MemoryClock, s.MaxMemoryClock
+func (conn PostgresConn) LatestData() (broadcast.Workstations, error) {
+	// pull all the machines in, then gpus for those machines
+	// has to all be done in a transaction to avoid race conditions
+	tx, err := conn.db.Begin()
```
Euphoride, line 318: With primarily Read queries, using a transaction is unnecessary and inefficient. There is no need to enforce read-write race condition freedom as we accept eventual consistency over to-the-instant consistency.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+
+		machine.LastSeen = time.Since(lastSeen)
+		machine.Gpus = nil
+
+		groups[*groupName] = append(groups[*groupName], machine)
+	}
+
+	// check for error whilst iterating, continuing if it's "no results"
+	err = machines.Err()
+	if err != nil && !errors.Is(sql.ErrNoRows, err) {
```
Euphoride: This is incorrect. The sql.ErrNoRows error is typically relevant when querying for a single row (e.g., with QueryRow). When iterating over multiple rows with Query, an empty result set doesn't result in an error but simply leads to no iterations of the loop.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
-			s.MemoryClock, s.MaxMemoryClock
+func (conn PostgresConn) LatestData() (broadcast.Workstations, error) {
+	// pull all the machines in, then gpus for those machines
+	// has to all be done in a transaction to avoid race conditions
+	tx, err := conn.db.Begin()
+	if err != nil {
+		return nil, err
+	}
+
+	groups := make(map[string][]broadcast.Workstation)
```
Euphoride, line 323: NIT: Preallocating the size of this is an efficiency boost.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+	machines, err := tx.Query(`SELECT GroupName, Hostname, CPU, Motherboard,
+		Notes, LastSeen
+		FROM Machines`)
+	if err != nil {
+		return nil, errors.Join(err, tx.Rollback())
+	}
+
+	for machines.Next() {
+		var groupName *string
```
Euphoride: Use default values in the SQL query instead rather than custom logic.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
 // TODO: consider returning workstationGroup
-func (conn PostgresConn) LatestData() ([]uplink.GpuStatsUpload, error) {
-	// we pull Uuid twice so we can put one into Stat sample and the other into Info
-	rows, err := conn.db.Query(`SELECT g.Machine, g.Uuid, g.Uuid, g.Name,
-			g.Brand, g.DriverVersion, g.MemoryTotal,
-			s.MemoryUtilisation, s.GpuUtilisation, s.MemoryUsed,
-			s.FanSpeed, s.Temp, s.MemoryTemp, s.GraphicsVoltage,
-			s.PowerDraw, s.GraphicsClock, s.MaxGraphicsClock,
-			s.MemoryClock, s.MaxMemoryClock
+func (conn PostgresConn) LatestData() (broadcast.Workstations, error) {
```
Euphoride, line 315: Isn't it possible here to use SQL's grouping mechanism to group things for you. SQL engines are **heavily** optimised for this work, why are we reinventing the wheel?

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+	}
+
+	return result, tx.Commit()
+}
+
+// get the latest stat for all the gpus on a machine
+func getGpus(host string, tx *sql.Tx) ([]broadcast.GPU, error) {
+	result := make([]broadcast.GPU, 0)
+
+	gpus, err := tx.Query(`SELECT g.Uuid, g.Name, g.Brand,
```
Euphoride, line 388: Better aliases on this query would be ideal.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+		result = append(result, gpu)
 	}
 
-	// flatten map structure
-	var result []uplink.GpuStatsUpload
-	for key, value := range latest {
-		result = append(result, uplink.GpuStatsUpload{Hostname: key,
-			GPUInfos: value.infos, Stats: value.stats})
+	err = gpus.Err()
+	if err == nil || errors.Is(err, sql.ErrNoRows) {
```
Euphoride: DITTO on `ErrNoRows`

internal/database/postgres.go
```diff
@@ -408,12 +469,12 @@ func (conn PostgresConn) RemoveMachine(machine broadcast.RemoveMachine) error {
...
 
 		if err != nil {
-			errors.Join(err, tx.Rollback())
+			return errors.Join(err, tx.Rollback())
 		}
 	}
 
 	err = rows.Err()
-	if !errors.Is(sql.ErrNoRows, rows.Err()) {
+	if err != nil && !errors.Is(sql.ErrNoRows, err) {
```
Euphoride: DITTO on ErrNoRows

Euphoride (dismissed):

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+		machine.LastSeen = time.Since(lastSeen)
+		machine.Gpus = nil
+
+		groups[*groupName] = append(groups[*groupName], machine)
+	}
+
+	// check for error whilst iterating, continuing if it's "no results"
+	err = machines.Err()
+	if err != nil && !errors.Is(sql.ErrNoRows, err) {
+		return nil, errors.Join(err, tx.Rollback())
```
Euphoride, line 356: NIT: Not ideal, pollutes the error returned if rollback also fails. Go wants simple error messages, and simple flow. I'd suggest logging the rollback error and returning `err` so that you can actually test the type/presence/etc of this error properly without a rollback failure getting in the way.

internal/database/postgres.go
```diff
@@ -309,66 +311,125 @@ FROM TempDownsampled;
...
+	}
+
+	// attach gpus to all machines
+	// can't be done in the previous loop because we can't be iterating
+	// through two queries at once
+	for group := range groups {
```
Euphoride, line 362: Bar me misunderstanding this code, this shouldn't have an inner loop that defers to a new SQL query. Rather, a clean, well-formatted and efficient SQL query should be constructed to defer this logic to the SQL Engine.

The SQL query execution planner has an almost certain chance of planning something more efficient than this, with much better performance.

internal/database/postgres.go
danielg0: This isn't a change I want to include in this pr, see #129

internal/database/inmemory.go
danielg0: I believe so, these side effects are exposed in the `Workstation` struct on a call to `LatestData`

NathanielB123 (approved):

internal/database/postgres.go
danielg0: Yep, fixed in 2f3d334

internal/database/postgres.go
danielg0: Fixed in 2f3d334

internal/database/postgres.go
danielg0: Fixed in 2f3d334

internal/database/postgres.go
danielg0: I disagree that this pollutes the error. `errors.Join()` returns a wrapped error. If you're testing for the presence/type of an error, you should be using [`errors.Is()`](https://pkg.go.dev/errors#Is), it checks all the errors wrapped and returns true if any of the errors matched the target error.

internal/database/postgres.go
danielg0: Just to be clear, the aliases you're referring to here are how `g` refers to the `Gpus` table and `s` refers to the `Stats` table?

internal/database/postgres.go
danielg0: Done in f369989

internal/database/postgres.go
danielg0: How would you suggest I get the number of groups in the database before we iterate through the returned rows - another SQL query before?

internal/database/postgres.go
danielg0: Unless we find that our database queries are being too slow, I'd like to treat this as a non-blocking style issue. I think the existing query in `getGpus` is already too long to understand quickly, and really don't want to combine that with joining the `machine` table and pulling out all of its fields too. I personally prefer shorter queries that are more decomposed, because I find them easier to understand, debug and modify.

internal/database/postgres.go
danielg0: I don't think a `GROUP BY` query, if that's what you're suggesting, would work for `LatestData`, because we don't want to aggregate the GPUs so there's one row per machine/per group

internal/database/postgres.go
aDotInTheVoid: Yeah, I agree that it's fine to have kinda sucky SQL queries for now. We can always change them later, and theirs no reason holding up integration on this.

internal/database/postgres.go
aDotInTheVoid: Eh, you could end up with some weird anomalies of machines/gpu's moving inside the same read. I don't think DB perf is a major concern, so I'm happy to merge this as is.

aDotInTheVoid:
LGTM, when conflicts resolved.

PR #144, opened 2024-02-15
femto: Don't pass `http.ResponseWriter` by pointer.

aDotInTheVoid:
Minor go style thing. All the stuff in `net/http` does it this way, and it makes the call and use sites clearer.

r? @Euphoride


Euphoride:
LGTM :)

Issue #73, opened 2024-01-29
Auto-restart satlites that haven't contacted up


aDotInTheVoid:
Either use user-systemd, or:

- See we've not seen a hearbeat
- But we can manage to ping the machine
- So SSH in, and start the satelite again

aDotInTheVoid:
Done in #108, but the git's fudged, so it's currently impossible to review.

Euphoride:
Done in #150, clean PR, waiting on review

Euphoride:
#150 merged

Issue #31, opened 2024-01-23
ci: More lints.


aDotInTheVoid:
- go vet
- copylocks
- maybe golangci-lint

PR #150, opened 2024-02-15
DRAFT: Automatic SSH Restart

Euphoride:
Implements #73


internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
+package database
```
aDotInTheVoid, line 1: Why's this in `package database`?

internal/config/server_configuration.go
```diff
@@ -1,15 +1,19 @@
 package config
 
 type Server struct {
-	GSPort int `toml:"groundstation_port"`
-	WAPort int `toml:"webapi_port"`
+	GSPort          int    `toml:"groundstation_port"`
+	WAPort          int    `toml:"webapi_port"`
+	DeathTimeout    int    `toml:"death_timeout"`
```
aDotInTheVoid: NIT: Why are these on server? So far it seems like `[server]` is for stuff related to just networking with other services, and other bits are in their own section (like `[auth]`, `[onboard]`, etc)

internal/config/server_configuration.go
```diff
@@ -1,15 +1,19 @@
...
+	MonitorInterval int    `toml:"monitor_interval"`
+	User            string `toml:"ssh_user"`
```
aDotInTheVoid: NIT: These are duplicated with the ssh config on `[onboard]`

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+	"golang.org/x/net/icmp"
+	"golang.org/x/net/ipv4"
+)
+
+type SSHConfig struct {
+	User    string
+	Keypath string
+}
+
+func MonitorForDeadMachines(interval int, database Database, timespanForDeath int, l *slog.Logger, s SSHConfig) error {
```
aDotInTheVoid: NIT: Should probably pass a `time.Duration` here for `interval` and `timespanForDeath`, so it's clearer what the units are.

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+	downsampleTicker := time.NewTicker(time.Duration(interval))
+
+	for t := range downsampleTicker.C {
+		err := monitor(database, t, timespanForDeath, l, s)
+
+		if err != nil {
+			return err
```
aDotInTheVoid: If we error when monitoring a single remote conection, we'll shut down the whole control server. This seams unideal.

For the HTTP stuff, if a single request errors, we'll log that, but keep the whole node alive. We should do a similar thing here.

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	}
+
+	return nil
+}
+
+func monitor(database Database, t time.Time, timespanForDeath int, l *slog.Logger, s SSHConfig) error {
+	last_seens, err := database.LastSeen()
```
aDotInTheVoid: NIT: This should use mixedCaps (ie `lastSeens`): -- https://go.dev/doc/effective_go#mixed-caps

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func monitor(database Database, t time.Time, timespanForDeath int, l *slog.Logger, s SSHConfig) error {
```
aDotInTheVoid: NIT: Kinda unclear what `t` represents here.

DITTO NIT: wrt `timespanForDeath` type

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if err != nil {
+		l.Error("Failed to run command on remote: %s", err)
+		return err
+	}
+
+	return nil
+}
+
+func ping(hostname string, l *slog.Logger) bool {
```
aDotInTheVoid: NIT: What does this returntype indicate? Probably a good idea to rename this to `hostIsPingable`, so it's clear at the callsite aswell.

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+	// Resolve the hostname to an IP address
+	ips, err := net.LookupIP(hostname)
+	if err != nil {
+		l.Debug("Error resolving hostname:", err)
+		return false
+	}
+	var ip net.IP
+	for _, i := range ips {
+		if i.To4() != nil {
```
aDotInTheVoid: NIT: Does this mean we're limited to ipv4 machines only? That's fine for now, as long as we have a `// TODO: Support ipv6` somewhere.

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+		},
+	}
+	binMsg, err := msg.Marshal(nil)
+	if err != nil {
+		l.Debug("Error marshaling message:", err)
+		return false
+	}
+
+	// Listen for ICMP replies
+	conn, err := icmp.ListenPacket("ip4:icmp", "0.0.0.0")
```
aDotInTheVoid: This requires ["privileged raw ICMP endpoints"](https://pkg.go.dev/golang.org/x/net@v0.21.0/icmp#ListenPacket). Do we have that as non-root users on lab machines?

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func sshRestart(remote string, l *slog.Logger, s SSHConfig) error {
+	user := s.User
+	keypath := s.Keypath
+
+	key, err := os.ReadFile(keypath)
+	if err != nil {
+		l.Error("Unable to read key: %v", err)
```
aDotInTheVoid: NIT: `log/slog` implements an idea called structured logging, where all attributes are key-value pairs. The correct way to do this is:

```go
l.Error("Unable to read ssh key", "keypath", keypath, "err", err)
```

Which will make something like

```
2024/01/20 14:300 ERROR Unable to read ssh key keypath=/foo/bar error=file not found
```

What you've done will cause

```
2024/01/20 14:300 ERROR Unable to read ssh key: %v !BADKEY=no such file
```

See https://go.dev/blog/slog and https://pkg.go.dev/golang.org/x/exp/slog#Logger.Log for details

---

Aside: we should lint against this. CC #31

internal/database/autocontact.go
```diff
@@ -0,0 +1,169 @@
...
+	defer client.Close()
+
+	sess, err := client.NewSession()
+	if err != nil {
+		l.Error("Failed to create session: %v", err)
+		return err
+	}
+	defer sess.Close()
+
+	_, err = sess.Output("nohup ./data/gpuctl/satellite")
```
aDotInTheVoid: This should be relative to `OnboardConf.DataDir`: 

https://github.com/gpuctl/gpuctl/blob/71ad89d9e52adfd0a5adb1a979e1de8e88021cbb/internal/webapi/server.go#L35

In addition, it should respect the convention of storing the logs, like onboard does.

https://github.com/gpuctl/gpuctl/blob/71ad89d9e52adfd0a5adb1a979e1de8e88021cbb/internal/onboard/onboard.go#L80-L81

It's probably a good idea to move this code from `database` to `onboard`, as that way we can share the ssh-start and ssh-restart functionality.

internal/database/autocontact_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.Error("Expected an error due to unparsable key, but got none")
+	} else {
+		t.Log("Received expected error:", err)
+	}
+}
+
+func TestPing(t *testing.T) {
+	logger := slog.Default()
+
+	success := ping("192.0.2.1", logger)
```
aDotInTheVoid: NIT: Can you add a comment that this is designated as non-routable by https://datatracker.ietf.org/doc/html/rfc5737

internal/database/autocontact_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if success {
+		t.Error("Expected ping to fail for a non-routable IP address, but it reported success")
+	} else {
+		t.Log("Ping processed as expected")
+	}
+}
+
```
aDotInTheVoid: This needs a test for the happy path, as it's fragile, and doesn't work. This should either go to loopback, or google.com.


```go
func TestPingHappy(t *testing.T) {
	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
		Level: slog.LevelDebug,
	}))

	success := ping("google.com", logger)

	if !success {
		t.Error("Failed to ping google")
	}
}
```

```
alona@Ashtabula:~/dev/gpuctl$ go test -v -run TestPingHappy ./internal/database/
=== RUN   TestPingHappy
time=2024-02-15T22:23:13.550Z level=DEBUG msg="Error listening for ICMP packets:" !BADKEY="listen ip4:icmp 0.0.0.0: socket: operation not permitted"
    autocontact_test.go:95: Failed to ping google
--- FAIL: TestPingHappy (0.02s)
FAIL
FAIL    github.com/gpuctl/gpuctl/internal/database      0.027s
FAIL
```

internal/database/autocontact_test.go
```diff
@@ -0,0 +1,152 @@
...
+func TestSSHRestart(t *testing.T) {
+	// Setup
+	logger := slog.Default()
+	sshConfig := SSHConfig{
+		User:    "testuser",
+		Keypath: "/path/to/your/test/private/key",
+	}
+
+	err := sshRestart("invalid.remote.address", logger, sshConfig)
+	if err == nil {
```
aDotInTheVoid: This needs to be much more specific, as currently all the ssh tests just want a non-nil error.

Take a look at `errors.Is` and `errors.As` to try to find the error you expect here.

cmd/control/control.go
```diff
@@ -80,6 +80,13 @@ func main() {
 	go func() {
 		errs <- database.DownsampleOverTime(conf.Database.DownsampleInterval, db)
 	}()
+	go func() {
+		c := database.SSHConfig{
```
aDotInTheVoid: We already do a load of work to get an ssh key for onboarding, we should just use that.

https://github.com/gpuctl/gpuctl/blob/71ad89d9e52adfd0a5adb1a979e1de8e88021cbb/cmd/control/control.go#L51-L60

This will require changing `MonitorForDeadMachines` to take an `ssh.Signer` (& friends), but that's a good change.

internal/database/autocontact.go
aDotInTheVoid: Why's this in `database`? `onboard` seems like a much nicer place for it.

aDotInTheVoid (commented):
Some random go nitpicks here, but some key things that need to be changed:

- the `ping` requires root privalege to run. I'd love to not have this requirement.
- the control server shouldn't shutdown when this errors
- when restarting a machine in the DB, we should look for the satellite binary where `onboard` placed it.

I'd like to test this manually e2e, to make sure it works

internal/database/autocontact_test.go
Euphoride: FYI I think the only reason this doesn't work because you don't have perms to make raw sockets, and so you should test with sudo. The new implementation, with shell-based pinging, passes this test.

internal/database/autocontact.go
Euphoride: Onboard is probably a bad place for it semantically. I'll put it in control methinks

aDotInTheVoid:
Blockers for merging:

- Gate ping happy path test on not being in CI.
- Use same command as onboard to start satellite over ssh
  - Respect `onboard.data_dir`
  - Place log files
- Don't use `%s` in `slog` functions

PR #151, opened 2024-02-15
launch-dev: Build go binaries

aDotInTheVoid:
This avoids the case where the binaries don't exist, or are out of date.


YRMYJASKA:
Sounds good, overlooked this when making the script lol.

PR #152, opened 2024-02-15
Adds new endpoint for duration delta from now since last seen

Euphoride:
Title, adds new endpoint for Last Updated frontend display.

### DIFF
--- a/internal/webapi/server.go
+++ b/internal/webapi/server.go
@@ -3,6 +3,7 @@ package webapi
 import (
 	"log/slog"
 	"net/http"
+	"time"
 
 	"github.com/gpuctl/gpuctl/internal/authentication"
 	"github.com/gpuctl/gpuctl/internal/broadcast"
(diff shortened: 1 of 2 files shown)


internal/webapi/server.go
```diff
@@ -172,6 +174,34 @@ func (a *Api) confirmAdmin(auth authentication.Authenticator[APIAuthCredientals]
...
+func (a *Api) durationDelta(r *http.Request, l *slog.Logger) (*femto.Response[[]DurationDeltas], error) {
+	latest, err := a.DB.LastSeen()
+
+	if err != nil {
+		return nil, err
+	}
+
+	var deltas []DurationDeltas
+
+	now := time.Now().Unix() / 1e9
```
aDotInTheVoid: What's this `1e9` doing? Needs a comment (or make it a named-const)

internal/webapi/server.go
```diff
@@ -172,6 +174,34 @@ func (a *Api) confirmAdmin(auth authentication.Authenticator[APIAuthCredientals]
 	return femto.Ok(UsernameReminder{Username: u})
 }
 
+type DurationDeltas struct {
```
aDotInTheVoid: This should live in `broadcast`, with the rest of the webapi<->frontend types.

internal/webapi/server.go
```diff
@@ -47,6 +48,7 @@ func NewServer(db database.Database, auth authentication.Authenticator[APIAuthCr
 
 	femto.OnGet(mux, "/api/stats/all", api.AllStatistics)
 	femto.OnGet(mux, "/api/stats/offline", api.HandleOfflineMachineRequest)
+	femto.OnGet(mux, "/api/stats/deltas", api.durationDelta)
```
aDotInTheVoid: NIT: I'm not sure this is a great name for this endpoint, as it's not super descriptive. How about something like `/api/stats/since_last_seen`?

aDotInTheVoid:
LGMT

PR #153, opened 2024-02-15
Add error message when login fails

sirjacobofglais:
Fix backend so that POST requests update the header, and return 401 in cases of auth failure

### DIFF
--- a/frontend/src/Components/SignIn.tsx
+++ b/frontend/src/Components/SignIn.tsx
@@ -1,9 +1,10 @@
 import { useState } from "react";
 import { Box, Button, Heading, Input, VStack } from "@chakra-ui/react";
 import { useAuth } from "../Providers/AuthProvider";
-
+import { Failure, VTag } from "../Utils/Utils";
+ 
 export const SignIn = () => {
-  const { login } = useAuth();
+  const { login, user : authName } = useAuth();
 
   const [username, setUsername] = useState("");
   const [password, setPassword] = useState("");
@@ -33,6 +34,7 @@ export const SignIn = () => {
           onChange={(e) => setPassword(e.target.value)}
           bgColor={"white"}
         ></Input>
+        { authName.tag == VTag.Failure && (authName as Failure).error.message == "Username or password was incorrect!" ? (<p> Username or passowrd incorrect</p>) : (<p> test</p>)}
         <Button
           bgColor={"white"}
           onClick={async () => {
--- a/frontend/src/Components/SignIn.tsx
+++ b/frontend/src/Components/SignIn.tsx
@@ -1,10 +1,10 @@
 import { useState } from "react";
-import { Box, Button, Heading, Input, VStack } from "@chakra-ui/react";
+import { Box, Button, Heading, Input, Text, VStack } from "@chakra-ui/react";
 import { useAuth } from "../Providers/AuthProvider";
 import { Failure, VTag } from "../Utils/Utils";
- 
+
 export const SignIn = () => {
-  const { login, user : authName } = useAuth();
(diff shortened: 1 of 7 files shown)


internal/webapi/server.go
```diff
@@ -109,6 +109,11 @@ func (a *Api) Authenticate(auth authentication.Authenticator[APIAuthCredientals]
 	// Check if credientals are correct
 	token, err := auth.CreateToken(packet)
 
+	if err == authentication.InvalidCredientalsError || err == authentication.NotAuthenticatedError {
+		slog.Info("HELLLLLLLLLLLLLLLO")
```
aDotInTheVoid: NIT: I don't think you ment to commit this.

internal/webapi/server.go
```diff
@@ -109,6 +109,11 @@ func (a *Api) Authenticate(auth authentication.Authenticator[APIAuthCredientals]
 	// Check if credientals are correct
 	token, err := auth.CreateToken(packet)
 
+	if err == authentication.InvalidCredientalsError || err == authentication.NotAuthenticatedError {
```
aDotInTheVoid: This should use [`errors.Is`](https://pkg.go.dev/errors#Is), to allow wrapped errors

internal/femto/femto.go
```diff
@@ -175,6 +175,7 @@ func doPost[T any](f *Femto, w http.ResponseWriter, r *http.Request, handle Post
 		http.SetCookie(w, &data.Cookies[i])
 	}
 
+	w.WriteHeader(data.Status)
```
aDotInTheVoid, line 178: Oof

internal/webapi/server.go
```diff
@@ -109,6 +109,11 @@ func (a *Api) Authenticate(auth authentication.Authenticator[APIAuthCredientals]
...
+		slog.Info("HELLLLLLLLLLLLLLLO")
+		return &femto.Response[types.Unit]{Status: http.StatusUnauthorized}, nil
```
aDotInTheVoid, line 114: Could you add a test for this behaviour?

aDotInTheVoid (commented):
I've not reviewed the frontend side, get nathaniel or mila to approve that before merging

internal/authentication/authentication.go
```diff
@@ -12,7 +12,7 @@ const TokenCookieName = "token"
 
 var (
 	NotAuthenticatedError   = errors.New("User does not have a valid authentication token")
-	InvalidCredientalsError = errors.New("Invalid credientals for creating an authentication token")
+	InvalidCredientalsError = errors.New("Invalid credentials for creating an authentication token")
```
danielg0: NIT: If we're fixing the spelling mistake in the error message, could we possibly fix it in the name of the error too?

internal/webapi/server.go
sirjacobofglais: I'm pretty sure there's already a test for this - put the test was using GET instead of POST so was passing. updating

EDIT: This is not true - adding the test now

frontend/src/Components/SignIn.tsx
```diff
@@ -33,6 +34,13 @@ export const SignIn = () => {
           onChange={(e) => setPassword(e.target.value)}
           bgColor={"white"}
         ></Input>
+        {authName.tag === VTag.Failure &&
```
NathanielB123: I think a `validatedElim` would be neater here (no `as` casting or needing to export `VTag`, which is kind of an impl detail).

NathanielB123 (changes requested):

NathanielB123 (approved):

PR #156, opened 2024-02-18
Made changes to inmem api to support new DB interface methods

Euphoride:
Currently blocked on #143 for the test suite, tests should be available there.


internal/database/inmemory.go
```diff
@@ -219,16 +219,68 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
 }
 
 func (m *inMemory) NewMachine(machine broadcast.NewMachine) error {
-	// TODO: add actual functionality. This was just to make the code compile
-	return errors.New("NOT IMPLEMENTED FOR IN-MEMORY DB")
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	info := uplink.GPUInfo{
```
YRMYJASKA: why do we put random information here?

Euphoride: Atm this is still a draft, given #143 has merged and the side-effects on this method are visible, this'll change

YRMYJASKA: ah okay, im marking it a draft properly

internal/database/inmemory.go
```diff
@@ -1,11 +1,8 @@
 package database
```
YRMYJASKA: can you enable the commented out test in `unit_tests.go`?

internal/database/inmemory.go
```diff
@@ -225,26 +215,23 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
-
 		if sample.Time < minTime {
 			minTime = sample.Time
 		}
-
+		// Addressing potential duplicate PID handling
 		for _, proc := range sample.RunningProcesses {
-			// TODO: What if multiple samples have a process with the same PID?
-			processesMap[proc.Pid] = proc
+			processesMap[proc.Pid] = proc // Overwrites if duplicate, ensuring only the latest is kept
```
aDotInTheVoid, line 227: NIT: I'm not sure i fully agree this is the right semantics, could we keep the TODO?

internal/database/inmemory.go
```diff
@@ -261,26 +248,88 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
 }
 
 func (m *inMemory) UpdateMachine(changes broadcast.ModifyMachine) error {
-	// TODO: add actual functionality. This was just to make the code compile
-	return errors.New("NOT IMPLEMENTED FOR IN-MEMORY DB")
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, exists := m.lastSeen[changes.Hostname]; !exists {
+		return fmt.Errorf("machine with hostname %s does not exist", changes.Hostname)
```
YRMYJASKA: we have errors for these in `interface.go` so would be good to use them.

internal/database/inmemory.go
```diff
@@ -261,26 +248,88 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
+	}
+
+	machine, exists := m.machines[changes.Hostname]
+	if !exists {
+		return fmt.Errorf("machine with hostname %s does not exist for update", changes.Hostname)
```
YRMYJASKA: ditto, I would just throw the same error as for the one above, and wrap it if need be.

internal/database/inmemory.go
```diff
@@ -261,26 +248,88 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
+	var uuidToRemove string
+	for uuid, info := range m.infos {
+		if info.host == machine.Hostname {
+			uuidToRemove = uuid
+			break
+		}
+	}
+
+	if uuidToRemove == "" {
+		return fmt.Errorf("machine with hostname %s does not exist", machine.Hostname)
```
YRMYJASKA: see above

internal/database/inmemory.go
```diff
@@ -225,26 +215,23 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
-
 	aggregatedProcesses := make([]uplink.GPUProcInfo, 0, len(processesMap))
 	for _, proc := range processesMap {
 		aggregatedProcesses = append(aggregatedProcesses, proc)
 	}
-	// Ensure deterministic order.
-	slices.SortFunc(aggregatedProcesses, func(a, b uplink.GPUProcInfo) int {
-		return cmp.Compare(a.Pid, b.Pid)
+	// Sorting to ensure deterministic order
+	sort.Slice(aggregatedProcesses, func(i, j int) bool {
```
aDotInTheVoid: NIT: It's entirely stylistic, but I prefer `SortFunc` here (and in general). It's more typesafe (taking `S ~[]E` vs `any`) and doesn't have the footgun of comparing via a different list. It's also apparently faster:

> Note: in many situations, the newer [slices.SortFunc](https://pkg.go.dev/slices#SortFunc) function is more ergonomic and runs faster.

<https://pkg.go.dev/sort#Slice>

internal/database/inmemory.go
```diff
@@ -261,26 +248,88 @@ func CalculateAverage(samples []uplink.GPUStatSample) uplink.GPUStatSample {
...
 }
 
 func (m *inMemory) NewMachine(machine broadcast.NewMachine) error {
-	// TODO: add actual functionality. This was just to make the code compile
-	return errors.New("NOT IMPLEMENTED FOR IN-MEMORY DB")
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if _, exists := m.lastSeen[machine.Hostname]; exists {
+		return fmt.Errorf("machine with hostname %s already exists", machine.Hostname)
```
YRMYJASKA: make this into an error that can be then used similarly in postgres?

aDotInTheVoid (commented):
LGTM, fix the nits if you want.

YRMYJASKA (commented):
lgtm, just errors could be homogenized to what is laid out in the interface and uncommenting one test case that should pass in `unit_test.go`

danielg0 (changes requested):
Now these functions are implemented, you can remove the test workarounds (eg. [here](https://github.com/gpuctl/gpuctl/blob/97f8f8e2b40dcd5a871e70cfcf0d783b3bbe2e60/internal/database/unit_test.go#L438)) from the unit tests that run on remove, etc.

danielg0:
Also, why are quite a few comments being removed. I get some of them (TODOs being answered), but some ([here](https://github.com/gpuctl/gpuctl/pull/156/files#diff-7d02d54911d88cf793a226d80809b2e336b102a3350abf366e16f00fa76a380dL75) and [here](https://github.com/gpuctl/gpuctl/pull/156/files#diff-7d02d54911d88cf793a226d80809b2e336b102a3350abf366e16f00fa76a380dL75) for instance)) seem to be like they're still relevant?

YRMYJASKA:
> Now these functions are implemented, you can remove the test workarounds (eg. [here](https://github.com/gpuctl/gpuctl/blob/97f8f8e2b40dcd5a871e70cfcf0d783b3bbe2e60/internal/database/unit_test.go#L438)) from the unit tests that run on remove, etc.

oh god, i didn't know we just _skipped_ errors coming from them!

internal/database/inmemory.go
Euphoride: Determined a broken test, not blocking

Euphoride:
@danielg0 Will need you to unblock this PR. Workaround is removed, and otherwise the test suite is now out of scope of this PR.

> Also, why are quite a few comments being removed. I get some of them (TODOs being answered), but some ([here](https://github.com/gpuctl/gpuctl/pull/156/files#diff-7d02d54911d88cf793a226d80809b2e336b102a3350abf366e16f00fa76a380dL75) and [here](https://github.com/gpuctl/gpuctl/pull/156/files#diff-7d02d54911d88cf793a226d80809b2e336b102a3350abf366e16f00fa76a380dL75) for instance)) seem to be like they're still relevant?

Those are both the same comment. That said, I will fix the nits and styles and merge this in on basis of LGTMs

internal/database/unit_test.go
```diff
@@ -394,14 +394,7 @@ func machineInfoUpdatesWork(t *testing.T, db database.Database) {
...
 
-	err = db.UpdateMachine(fakeChange)
-
-	// TODO: remove this
-	// skip errors, as inmemory doesn't currently implement update
-	//assert.NoError(t, err)
-	if err != nil {
-		t.Skipf("Skipping with error, was %v", err)
-	}
+	_ = db.UpdateMachine(fakeChange)
```
danielg0: Don't we want to assert that no error is returned here, rather than discard the return value?

internal/database/unit_test.go
```diff
@@ -435,12 +428,7 @@ func removingMachine(t *testing.T, db database.Database) {
 		t.Error("Didn't find machine when we expected to")
 	}
 
-	// TODO: remove this
-	// workaround for inmem not implementing removal
 	err = db.RemoveMachine(broadcast.RemoveMachine{Hostname: fakeHost})
```
danielg0, line 432: Same here about asserting no error occurs

danielg0 (commented):
.

danielg0 (approved):

PR #159, opened 2024-02-18
Consolidate ssh code into `onboard`.

aDotInTheVoid:
Followup to #150

Advantages:

- Less code
- Single config type for all SSH stuff
- Only one place now needed to change HostKeyCallback
- Only one place where the `nohup` command is
- SSH session creation code consolidated

Disadvantages:

- The `onboard` name is now misleading.

### DIFF
--- a/cmd/control/control.go
+++ b/cmd/control/control.go
@@ -13,6 +13,7 @@ import (
 	"github.com/gpuctl/gpuctl/internal/config"
 	"github.com/gpuctl/gpuctl/internal/database"
 	"github.com/gpuctl/gpuctl/internal/groundstation"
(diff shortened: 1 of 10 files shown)


Euphoride:
Can we rename onboard to something better then given the name is now not applicable?

aDotInTheVoid:
In a followup!

Euphoride:
Not super happy approving a merge yet without naming changes given no reason the naming shouldn't be reflected now rather than later.

aDotInTheVoid:
Gonna rename it to `tunnel`, in a followup PR this evening, but going to merge as-is, so each PR only does one thing.

PR #162, opened 2024-02-19
Round all numerical figures to 0dp on frontend

sirjacobofglais:


NathanielB123:
I feel like rounding earlier would look superficially neater, but would be conceptually weird (rounding ultimately is to do with display of the data) so yeah this is probably fine.

PR #163, opened 2024-02-19
Allow signin via enter key in input boxes

sirjacobofglais:
Calls the login function when the enter key is pressed when focus is on username or password input.


NathanielB123:
LGTM!

PR #164, opened 2024-02-19
Implement "last seen" warnings on the front-end

NathanielB123:


NathanielB123:
Untested because last seen stuff only works for Postgres DB right now, but I think this should be more-or-less done

frontend/src/Components/WorkstationCardMinimal.tsx
```diff
@@ -1,40 +1,80 @@
...
+        {lastSeen !== undefined && lastSeen > LAST_SEEN_WARN_THRESH ? (
+          <Alert
+            roundedTopLeft="md"
+            roundedTopRight="md"
+            textAlign="center"
+            status="warning"
+          >
+            <AlertIcon></AlertIcon>
+            {lastSeen === undefined
+              ? "Missing provenance data!"
```
sirjacobofglais, line 43: nit: I think we should look at what the best words are to use here (maybe just take from what they have now?)

sirjacobofglais (approved):

PR #166, opened 2024-02-19
Send user information satellite -> groundstation.

### LINKED ISSUE #165: Send proc and user information satellite -> gs
title.

YRMYJASKA:
We sent proc info before, nothing about the users. Now we have functionality to send the full name (or username) to the gs.

Closed #165

### DIFF
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -185,8 +185,11 @@ func setGPUHandler(log *slog.Logger, isFakeGPUs bool) gpustats.GPUDataSource {
 		return gpustats.FakeGPU{}
 	} else {
 		passwdfile, err := os.Open("/etc/passwd")
-		passwd, err1 := passwd.Parse(passwdfile)
-		err = errors.Join(err, err1)
+		if err != nil {
+			log.Error("Could not open passwd file, will not be able to report users' names", "err", err)
+			return gpustats.NvidiaGPUHandler{}
+		}
+		passwd, err := passwd.Parse(passwdfile)
 		if err != nil {
 			log.Error("Could not read passwd file, will not be able to report users' names", "err", err)
 			return gpustats.NvidiaGPUHandler{}
--- a/cmd/satellite/main.go
+++ b/cmd/satellite/main.go
@@ -11,6 +11,8 @@ import (
 	"github.com/gpuctl/gpuctl/internal/config"
 	"github.com/gpuctl/gpuctl/internal/femto"
 	"github.com/gpuctl/gpuctl/internal/gpustats"
+	"github.com/gpuctl/gpuctl/internal/passwd"
+	"github.com/gpuctl/gpuctl/internal/procinfo"
 	"github.com/gpuctl/gpuctl/internal/uplink"
 )
 
@@ -49,7 +51,7 @@ func main() {
 		hostname: host,
 	}
 
-	hndlr := setGPUHandler(satellite_configuration.Satellite.FakeGPU)
+	hndlr := setGPUHandler(log, satellite_configuration.Satellite.FakeGPU)
 
 	// Send initial infopacket of GPUInfo
 	log.Info("Sending initial GPU context")
@@ -178,11 +180,19 @@ func processStats(stats [][]uplink.GPUStatSample) []uplink.GPUStatSample {
 	return stats[len(stats)-1]
 }
 
-func setGPUHandler(isFakeGPUs bool) gpustats.GPUDataSource {
+func setGPUHandler(log *slog.Logger, isFakeGPUs bool) gpustats.GPUDataSource {
 	if isFakeGPUs {
 		return gpustats.FakeGPU{}
 	} else {
-		return gpustats.NvidiaGPUHandler{}
+		passwdfile, err := os.Open("/etc/passwd")
+		passwd, err1 := passwd.Parse(passwdfile)
+		err = errors.Join(err, err1)
+		if err != nil {
+			log.Error("Could not read passwd file, will not be able to report users' names", "err", err)
+			return gpustats.NvidiaGPUHandler{}
+		}
+
+		return gpustats.NvidiaGPUHandler{Lookup: procinfo.PasswdToLookup(passwd)}
 	}
 }
 
--- a/internal/gpustats/nvidia.go
+++ b/internal/gpustats/nvidia.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/gpuctl/gpuctl/internal/procinfo"
 	"github.com/gpuctl/gpuctl/internal/uplink"
 )
 
@@ -293,6 +294,7 @@ func (xml NvidiaSmiLog) ExtractGPUInfo() ([]uplink.GPUInfo, error) {
 func (xml NvidiaSmiLog) ExtractGPUStatSample() ([]uplink.GPUStatSample, error) {
 	var res []uplink.GPUStatSample
 
+	// Go over gpus in xml
 	for _, gpu := range xml.Gpu {
 		// TODO: add the new information as well
 		fanSpeed, err := parseFloatWithUnit(gpu.FanSpeed, "fan speed")
@@ -408,7 +410,7 @@ func ParseNvidiaSmi(input []byte) (NvidiaSmiLog, error) {
 }
 
 // Get the Nvidia GPU status directly from the computer using `nvidia-smi`
-func GetNvidiaGPUStatus() (NvidiaSmiLog, error) {
+func getNvidiaGPUStatus() (NvidiaSmiLog, error) {
 	output, err := exec.Command("nvidia-smi", "-q", "-x").Output()
 	if err != nil {
 		return NvidiaSmiLog{}, err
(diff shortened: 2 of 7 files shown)


internal/procinfo/pidtouser.go
```diff
@@ -0,0 +1,51 @@
...
+	// get uid
+	filename := "/proc/" + strconv.FormatUint(pid, 10)
+	statcall, err := os.Stat(filename)
+	if err != nil {
+		return zero, err
+	}
+	stat, okay := statcall.Sys().(*syscall.Stat_t)
+	if okay {
+		return "", err
+	}
```
danielg0, line 28: This `if` statement confuses me a little bit. I think that we're casting the file info to something OS-specific, and whether that succeeds is written to `okay`? If so, why do we return an empty name if we did cast successfully?

Secondly, I believe that the returned `err` in this case is always nil, as we've just compared it to nil and not changed it after. I think we instead make a new "Couldn't cast" error and return that

internal/gpustats/nvidia.go
```diff
@@ -417,21 +419,42 @@ func GetNvidiaGPUStatus() (NvidiaSmiLog, error) {
...
+// Struct to act as our adapter
+type NvidiaGPUHandler struct {
+	lookup procinfo.UidLookup
+}
+
+func populateNames(samples *[]uplink.GPUStatSample, lookup procinfo.UidLookup) {
+	for i, sample := range *samples {
+		for j, proc := range sample.RunningProcesses {
+			name, err := lookup.Get(proc.Pid)
+			// Intentionally leave unresolved names as zero, don't want to fail
```
danielg0, line 431: Not failing is fine, but could we possibly still log the error so we're aware issues are occurring?

danielg0:
Just tried on `gpu07.doc.ic.ac.uk` and it doesn't seem to be giving back any names

internal/gpustats/nvidia_test.go
```diff
@@ -138,3 +138,7 @@ func TestNvidiaSmiInvalidDataParse(t *testing.T) {
 		}
 	}
 }
+
+func TestNvidiaCorrectNamePopulation(t *testing.T) {
+
```
aDotInTheVoid: It looks like you wanted to have something here??

internal/procinfo/pidtouser.go
```diff
@@ -0,0 +1,51 @@
...
+	"syscall"
+)
+
+var (
+	ErrorNoUserFound = errors.New("Could not find user that owns the given process from passwd file")
+)
+
+type UidLookup map[uint32]string
+
+func (lookup UidLookup) Get(pid uint64) (string, error) {
```
aDotInTheVoid: NIT: This name doesn't make sense without knowing the receiver. Could you rename it to something like `UsernameForPid`?

internal/procinfo/pidtouser.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+	return name, nil
+}
+
+func PasswdToLookup(entries passwd.Passwd) UidLookup {
+	lookup := make(UidLookup)
+	for _, entry := range entries {
+		// Try to assign full name to lookup, otherwise use username
+		fullname := entry.ParseGecos().FullName
+		if len(fullname) < 1 {
```
aDotInTheVoid: This should just be `fullname == ""`. Theirs no reason to be so obtuse here. The equality check is much clearer.

internal/procinfo/pidtouser_test.go
```diff
@@ -0,0 +1,30 @@
...
+var ashtabula string
+
+func loadAshtabula(t *testing.T) passwd.Passwd {
+	t.Helper()
+	users, err := passwd.Parse(strings.NewReader(ashtabula))
+	require.NoError(t, err)
+	return users
+}
+
+func TestPasswdToLookup(t *testing.T) {
```
aDotInTheVoid, line 26: NIT: Can call [`t.Parallel`](https://pkg.go.dev/testing#T.Parallel) here

internal/uplink/gpustats.go
```diff
@@ -41,4 +41,5 @@ type GPUProcInfo struct {
 	Pid     uint64  `json:"pid"`
 	Name    string  `json:"name"`
 	MemUsed float64 `json:"used_memory"`
+	Owner   string  `json:"owner"`
```
aDotInTheVoid: NIT: It's worth adding a comment here about `Owner` being the name of the user running the job, in human readable form.

internal/procinfo/pidtouser.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+var (
+	ErrorNoUserFound = errors.New("Could not find user that owns the given process from passwd file")
+)
+
+type UidLookup map[uint32]string
+
+func (lookup UidLookup) Get(pid uint64) (string, error) {
+	var zero string
```
aDotInTheVoid: NIT: Given that `""` is literally twice as short as `zero`, I'd rather we used that.

internal/procinfo/pidtouser.go
aDotInTheVoid: Yeah, this is wildly broken.

I think the correct code looks like

```go
if !ok {
    return "", ErrUnsuportedOs
}
```

NIT: Also go uses `ok` (not `okay`) for this sort of variable by convention.  grep for `, ok` in [effective go](https://go.dev/doc/effective_go)

internal/gpustats/nvidia.go
```diff
@@ -417,21 +419,42 @@ func GetNvidiaGPUStatus() (NvidiaSmiLog, error) {
...
 }
 
-// Dummy struct to act as our adapter
-type NvidiaGPUHandler struct{}
+// Struct to act as our adapter
+type NvidiaGPUHandler struct {
+	lookup procinfo.UidLookup
+}
+
+func populateNames(samples *[]uplink.GPUStatSample, lookup procinfo.UidLookup) {
```
aDotInTheVoid: NIT: You can just take `samples []uplink.GPUStatsSample`, instead of needing a pointer here, as you never modify the slice itself (only it's contents)

internal/procinfo/pidtouser_test.go
```diff
@@ -0,0 +1,30 @@
...
+//go:embed testdata/ashtabula
+var ashtabula string
+
+func loadAshtabula(t *testing.T) passwd.Passwd {
+	t.Helper()
+	users, err := passwd.Parse(strings.NewReader(ashtabula))
+	require.NoError(t, err)
+	return users
+}
+
```
aDotInTheVoid, line 25: Could you add a test that calls `.Get`/`.PidToUser` on [os.Getpid](https://pkg.go.dev/os#Getpid)? (Probably with a map that sets this PID to a preset value, rather than passing /etc/passwd)

internal/procinfo/pidtouser.go
YRMYJASKA: Noted, was used to writing checks with `err != nil`, hence the inversion of logic here

internal/gpustats/nvidia.go
YRMYJASKA: oh what? good to know that

internal/gpustats/nvidia.go
YRMYJASKA: then we would need to thread the needle with getting a logger down to this level. I dont think that is worth it. There might be other solutions but I rather not.

internal/gpustats/nvidia_test.go
YRMYJASKA: Had something in mind, then promptly forgot about it. I really should have looked over this pr more before i requested review

YRMYJASKA:
> Just tried on `gpu07.doc.ic.ac.uk` and it doesn't seem to be giving back any names

~~are you sure? I just tried this with my computer and it did send process information just fine?~~ EDIT: yeah, might not have worked because of my extremely dumb code before. Now it should work (I tested it out)

cmd/satellite/main.go
```diff
@@ -178,11 +180,19 @@ func processStats(stats [][]uplink.GPUStatSample) []uplink.GPUStatSample {
...
 }
 
-func setGPUHandler(isFakeGPUs bool) gpustats.GPUDataSource {
+func setGPUHandler(log *slog.Logger, isFakeGPUs bool) gpustats.GPUDataSource {
 	if isFakeGPUs {
 		return gpustats.FakeGPU{}
 	} else {
-		return gpustats.NvidiaGPUHandler{}
+		passwdfile, err := os.Open("/etc/passwd")
+		passwd, err1 := passwd.Parse(passwdfile)
```
aDotInTheVoid: If this open fails, we should bail early, as `passwdfile` may be a `nil` `io.Reader`, and we'll die

internal/gpustats/nvidia_test.go
aDotInTheVoid: I think you need some sort of test for PopulateNames here.

internal/gpustats/nvidia.go
danielg0: Fair enough, feel free to merge without.
Has everyone else been using logger objects and passing them around? I've mainly just been using `slog.Info(...)` functions straight from the package

YRMYJASKA: hmmm, you have a point there. I've been mostly using logger objects

aDotInTheVoid: > Has everyone else been using logger objects and passing them around?

It's super usefull on the control, as it allows associating log entries with the request that resulted in them.

Not sure if it matters here as much on satellite,

aDotInTheVoid:
LGTM, unless you can figure out a test for the `PopulateNames` stuff

danielg0 (approved):
LGTM

PR #178, opened 2024-02-20
procinfo: Use `golang.org/x/sys/` instead of `syscall`

aDotInTheVoid:
From the `syscall` docs:

> NOTE: Most of the functions, types, and constants defined in this package are
> also available in the golang.org/x/sys package. That package has more system
> call support than this one, and most new code should prefer that package where
> possible. See https://golang.org/s/go1.4-syscall for more information.

More concretely, we don't need a typecast on the result, which is nice.


YRMYJASKA (commented):
LGTM

danielg0 (approved):
LGTM

PR #183, opened 2024-02-20
DRAFT: Admin Page Refactor

Euphoride:
A short refactor seperating the admin page into manageable chunks.

CC @NathanielB123


frontend/src/Hooks/hooks.tsx
```diff
@@ -0,0 +1,51 @@
+import { STATS_PATH } from "../Config/Paths";
```
NathanielB123: We already have a file for hooks (`Utils/Hooks.tsx` - but those are general purpose, not domain specific like the ones here, so having a new file seems fine.
NIT: Name of the file should probably be capitalised to be consistent tho ;)

frontend/src/Pages/AdminPanel.tsx
```diff
@@ -1,242 +1,20 @@
...
-                    </Tr>
-                  );
-                });
-              }),
-            )}
-          </Tbody>
-        </Table>
-      </TableContainer>
+      <AddMachineForm groups={groups} />
+      <GroupInfoManagement groups={groups} />
```
NathanielB123, line 17: Somehow the layout of the page got changed quite significantly by abstracting this into a new component (the table is much wider than it was), but I don't think I have a preference either way, so happy to keep the change.

frontend/src/Components/EditableFields.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { useModifyInfo } from "../Hooks/hooks";
+import { WorkStationData } from "../Data";
+
+type EditableFieldProps = {
+  workstation: WorkStationData;
+  fieldKey: "cpu" | "motherboard" | "notes" | "group";
+  placeholder: string;
+};
+
+export function EditableField({
```
NathanielB123: So much nicer, thanks!
Tho, NIT: Everywhere else I have been using `export const EditableField = (...) => ...` syntax instead of `function`, so could change to use that to be consistent

NathanielB123 (approved):

NathanielB123:
Temporarily reverting so we can merge the other PRs first

PR #184, opened 2024-02-20
webapi file upload endpoint

### LINKED ISSUE #186: Attaching files endpoint and DB functionality
This issue is a part of the wider issue #21. This needs to be completed before.

YRMYJASKA:
Closes #186


internal/webapi/server.go
```diff
@@ -70,6 +71,10 @@ func (a *Api) AllStatistics(r *http.Request, l *slog.Logger) (*femto.Response[br
 	if err != nil {
 		return nil, err
 	}
+	if data == nil {
```
YRMYJASKA, line 79: this bit is the important bit for returning `[]` from allstats instead of `null` when the db is empty.

internal/database/postgres.go
```diff
@@ -76,6 +76,17 @@ func createTables(db *sql.DB) error {
 		return err
 	}
 
+	_, err = db.Exec(`CREATE TABLE IF NOT EXISTS Files (
```
danielg0, line 79: I don't think you've specified the primary key of this table. I'd suggest making a composite key of `(Hostname, Filename)`

internal/database/postgres.go
```diff
@@ -622,3 +634,99 @@ func (conn PostgresConn) LastSeen() ([]broadcast.WorkstationSeen, error) {
 
 	return seens, nil
 }
+
+func (conn PostgresConn) AttachFile(attach broadcast.AttachFile) error {
+	// Make sure we delete duplicates. Somewhat hacky
```
danielg0: If we want uploading a file with a duplicate name to replace the old one (which I'm fine with, so long as it's very clear that's what's happening on the frontend), it may be cleaner to use an `ON CONFLICT` clause in the INSERT statement (see [here](https://github.com/gpuctl/gpuctl/blob/43d53a8b7393819cd3ec3167b664c3889fcf33eb/internal/database/postgres.go#L210) for an example).

This would be a nit, but this select statement isn't in a transaction with the insert, so I think there's a race condition here.

internal/database/postgres.go
```diff
@@ -622,3 +634,99 @@ func (conn PostgresConn) LastSeen() ([]broadcast.WorkstationSeen, error) {
...
+	if err != nil {
+		return ErrNoSuchMachine
+	}
+	return nil
+}
+
+func (conn PostgresConn) GetFile(hostname string, filename string) (broadcast.AttachFile, error) {
+	file := broadcast.AttachFile{Hostname: hostname, Filename: filename}
+
+	rows, err := conn.db.Query(`SELECT Mime, File
```
danielg0: NIT: If `(Hostname, Filename)` is the primary key of the Files table, this can be QueryRow rather than Query

internal/database/postgres.go
danielg0, line 1: I believe `RemoveMachine()` needs to be updated so that all a machine's files are detected before its rows in the machine table is removed

danielg0: ^ adding a file to a machine in the [unit test for removing](https://github.com/gpuctl/gpuctl/blob/43d53a8b7393819cd3ec3167b664c3889fcf33eb/internal/database/unit_test.go#L423) would be great too

danielg0 (commented):
Here's a couple comments on the database code. Webapi stuff lgtm

internal/database/postgres.go
YRMYJASKA: Note to myself: look above for example

internal/database/postgres.go
YRMYJASKA: this needs inmemoryd db remove machine to work as well. Wrote test, just uncomment when possible

YRMYJASKA: Depends on #156 being merged for us to enable this unit test

PR #187, opened 2024-02-22
Sort data table by isFree

sirjacobofglais:
Testing on dev is painful because we don't have satellites set up - so can we just test on prod? 🥺


NathanielB123:
Closing in favour of https://github.com/gpuctl/gpuctl/pull/187

PR #188, opened 2024-02-22
Functionality for rebooting machines command being put into clipboard

Euphoride:
Title.


NathanielB123:
Once https://github.com/gpuctl/gpuctl/pull/194 lands on trunk, I would heavily recommend merging trunk in rather than rebasing (solving the conflicts for the revert followed by the re-revert sounds like it would be a nightmare)

NathanielB123 (approved):
LGTM

PR #191, opened 2024-02-22
got rid of some suspicious duds in db queries and switched to 8.8.8.8…

YRMYJASKA:
… instead of google for no particular reason :^)


danielg0 (approved):
LGTM

PR #192, opened 2024-02-22
get rid of the last seen endpoint as we dont need it

YRMYJASKA:
This information is now being passed in getting all stats. We also need to fix this on the front-end @NathanielB123


danielg0 (approved):
LGTM. We don't need the `DurationDeltas` struct anymore, so I'm going to merge after removing that

PR #193, opened 2024-02-22
Deployment fixes

danielg0:
A couple of small fixes to deployment we need to get site working

### DIFF
--- a/deploy/control.prod.toml
+++ b/deploy/control.prod.toml
@@ -30,5 +30,5 @@ port = 443
 # intervals in nanoseconds
 data_interval = 15000000000
 heartbeat_interval = 5000000000
-fake_gpu = true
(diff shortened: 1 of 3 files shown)


NathanielB123 (approved):
LGTM

deploy/control.prod.toml
```diff
@@ -16,7 +27,8 @@ hostname = "gpuctl.perial.co.uk"
 port = 443
 
 [Onboard.Remote.Satellite]
+# intervals in nanoseconds
```
aDotInTheVoid, line 30: Ooof that some of these are ns and some are secs. I should land 1ff7eccde66bbf36bf63e012e245ff9685b483a9, so we can write `foo = "1s"`

PR #195, opened 2024-02-22
Remove Last Seen

NathanielB123:
Instead of using the "last seen" endpoint, we now use the data given to us by "all stats"


danielg0 (approved):
LGTM

PR #202, opened 2024-02-23
`RemoveMachine` postgres implementation fixes

danielg0:
When we have a set of rows open from a previous query, we **cannot** start a new query


NathanielB123 (approved):
LGTM

PR #204, opened 2024-02-23
Add the username to the onboarding data directory

danielg0:
Stops issues where we try to onboard a machine that has already been onboarded in a previous deployment with a different user's ssh key (we don't have permission to delete the other users install)


NathanielB123 (approved):
LGTM!

PR #208, opened 2024-02-23
Specify Content-Type when doing manual return

danielg0:
Because content type header was missing, femto was trying to coerce broadcast.Workstations to a binary and failing, due to the variable-length fields.

### DIFF
--- a/internal/webapi/server.go
+++ b/internal/webapi/server.go
@@ -76,11 +76,7 @@ func (a *Api) AllStatistics(r *http.Request, l *slog.Logger) (*femto.Response[br
 
 	if data == nil {
 		// dont just return nil, which would not be marshalled properly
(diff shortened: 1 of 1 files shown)


internal/webapi/server.go
```diff
@@ -76,7 +76,11 @@ func (a *Api) AllStatistics(r *http.Request, l *slog.Logger) (*femto.Response[br
 
 	if data == nil {
 		// dont just return nil, which would not be marshalled properly
-		return &femto.Response[broadcast.Workstations]{Status: http.StatusOK, Body: broadcast.Workstations{}}, nil
+		return &femto.Response[broadcast.Workstations]{
```
YRMYJASKA: This could be `femto.Ok()` that already sets the content type to `json` by default.

YRMYJASKA (approved):

PR #224, opened 2024-02-28
Add spinner and toast for addMachine response

sirjacobofglais:


frontend/src/Components/AddMachineForm.tsx
```diff
@@ -37,11 +43,26 @@ export const AddMachineForm = ({
         <Button
           w="5%"
           onClick={() => {
+            setSpinner(true);
+            onNextFetch((stats) => {
```
NathanielB123: In the case the first `fetch` does not include the new machine, this won't work as we only fire `onNextFetch` callbacks once.

NathanielB123: Partially fixed in https://github.com/gpuctl/gpuctl/pull/224/commits/7149ab7deadc41ea2458d82e976ae24fc678f907, fully fixed in https://github.com/gpuctl/gpuctl/pull/224/commits/1f5f87568e3e391987cbd8ef78a6f0730bcdfdef

PR #225, opened 2024-02-28
Remove production build of frontend from CI

NathanielB123:


.github/workflows/ci.yaml
```diff
@@ -71,13 +71,13 @@ jobs:
...
         run: test -z $(gofmt -l .)
-      
+
       #
       # Frontend
       #
-      - name: Build Frontend
-        run: npm run build
-        working-directory: ./frontend
+      # - name: Build Frontend
```
sirjacobofglais: remove pls

sirjacobofglais (dismissed):

.github/workflows/ci.yaml
NathanielB123: Fiiiiiiine https://github.com/gpuctl/gpuctl/pull/225/commits/5109ed7c123bc9a3d1e04f4312fbee2ac7e8e1a4

aDotInTheVoid:
For future reference, this is OK because we do this in the docker build

PR #229, opened 2024-02-29
Add toasts for removing machine

sirjacobofglais:


frontend/src/Providers/AuthProvider.tsx
```diff
@@ -156,8 +156,8 @@ export const AuthProvider = ({
 
       fire(async () => {
         const r = await authFetch(path, init);
-        if (!r.ok) {
-          if (r.status === 401) logout();
+        if (r.status == 401) {
```
NathanielB123: This should be `===` I think?

NathanielB123 (approved):

frontend/src/Providers/AuthProvider.tsx
NathanielB123: https://github.com/gpuctl/gpuctl/pull/229/commits/c6a4f64e8d53e55483756da4b0015d5891502ad7

Issue #219, opened 2024-02-27
Fix downsampling crashes


sirjacobofglais:

aDotInTheVoid:
Any logs of this happening?

NathanielB123:
> Any logs of this happening?

https://discord.com/channels/1193956842368417863/1193956843333103758/1210970497542529124
https://discord.com/channels/1193956842368417863/1193956843333103758/1211469576089374832

NathanielB123:
I think it would also be ideal to sanity check the results of downsampling - i.e. just run for a while and see if the values are in reasonable ranges, because right now I don't think we have any front-end behaviour that allows us to see older/downsampled data.

danielg0:
> I think it would also be ideal to sanity check the results of downsampling - i.e. just run for a while and see if the values are in reasonable ranges, because right now I don't think we have any front-end behaviour that allows us to see older/downsampled data.

Yes, I agree, I don't know if this is worth properly fixing til we have long term data stats coming out of the db to compare with in unit tests. I'd suggest we just make it so errors are logged rather than returned from downsample, then fix "properly" once we have those long term data stats coming out of the db

PR #230, opened 2024-02-29
Improve logging of errors in production

danielg0:
- Log the source of an error when one occurs in `control`
- Log errors that occur during down-sampling rather than crashing (helps with #219 in the short term)


NathanielB123 (approved):
LGTM

PR #231, opened 2024-02-29
Use keys based on name and pick button colour

sirjacobofglais:


NathanielB123 (approved):
`isEven` probably isn't the best name but ¯\\\_(ツ)\_/¯

Issue #223, opened 2024-02-28
Add owner field on frontend + db


sirjacobofglais:

danielg0:
Going to add an extra text field to represent this

PR #232, opened 2024-02-29
Track machine owner in the database.

danielg0:
Part of #223

No changes on webapi should be needed, iirc. Just changing frontend to use owner


YRMYJASKA (approved):
LGTM

PR #233, opened 2024-03-03
Fix date we test downsampling

danielg0:
The precise number of downsamples we get seems to vary depending on when the test is run (it's started failing since the start of the month.

Therefore, as a temporary patch, fix the rundate of the downsample test to be the beginning of Febuary, this should cover us for now.


NathanielB123 (approved):
LGTM

PR #235, opened 2024-03-04
Fixed adding machines to db

### LINKED ISSUE #214: Fix WebAPI/backend not adding onboarded machines to specified groups

YRMYJASKA:
* corrected semantics of updating (should not add machines, ever)
* merged non-used NewMachine and createMachine (createMachine is more general)
* made api endpoint for adding machines to use NewMachine
* Added tests for expected behaviour
* fixed inmemory implementation to conform to these

Closes #214

### DIFF
--- a/internal/database/inmemory.go
+++ b/internal/database/inmemory.go
@@ -326,7 +326,8 @@ func (m *inMemory) UpdateMachine(changes broadcast.ModifyMachine) error {
 	defer m.mu.Unlock()
 
 	if _, exists := m.lastSeen[changes.Hostname]; !exists {
-		return ErrMachineNotPresent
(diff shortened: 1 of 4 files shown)


danielg0 (approved):
LGTM!

NathanielB123:
I imagine this changes front-end experience a bit. Before this PR we would get a toast/see the new machine after the backend received a packet from the satellite - am I right in assuming that now the machine will show up in the database before that packet arrives? Is the idea that it just first shows up with no GPUs?

YRMYJASKA:
yeah, now it will first add the machine as a machine without any information on it. But you made me realise that there is now a subtle race-condition: if the satellite packet arrives first, then the machine will be added into the `Shared` group

PR #239, opened 2024-03-04
graphs functionality on the backend

YRMYJASKA:


internal/broadcast/types.go
```diff
@@ -91,3 +91,15 @@ type WorkstationSeen struct {
...
+type HistoricalData []HistoricalDataPoint
+
+type HistoricalDataPoint struct {
+	Timestamp int64 `json:"timestamp"`
+	Samples   []GPU `json:"sample"`
+}
+
+type AggregateData struct {
+	PercentUsed int    `json:"percent_used"`
+	TotalEnergy uint64 `json:"total_energy"`
```
danielg0: Would be great to get a quick description/units we transfer energy in

internal/database/postgres.go
```diff
@@ -720,3 +720,148 @@ func (conn PostgresConn) RemoveFile(remove broadcast.RemoveFile) error {
...
+		s.GraphicsClock,
+		s.MaxGraphicsClock,
+		s.MemoryClock,
+		s.MaxMemoryClock,
+		s.InUse,
+		s.UserName
+		FROM Stats s
+		INNER JOIN GPUs g ON g.Uuid = s.Gpu
+		WHERE g.Machine=$1
+		ORDER BY s.Received`,
```
danielg0: For a machine with more than one GPU, how are we enforcing an ordering that alternates between GPUs in the same pattern? If I'm reading this right, I think we could potentially get the data out the database in the order `GPU1 @ 1s, GPU2 @ 1s, GPU2 @ 2s, GPU1 @ 2s, ...`.

It should be enough to make this line `ORDER BY s.Recieved, s.Gpu`

internal/database/postgres.go
```diff
@@ -720,3 +720,148 @@ func (conn PostgresConn) RemoveFile(remove broadcast.RemoveFile) error {
...
+		WHERE s.Received > $1
+		ORDER BY s.Received`,
+		threshold,
+	)
+	if err != nil {
+		return data, err
+	}
+
+	// XXX: I hate EVERYTHING about the rest of this function,
+	//  	but idk how to write a better SQL query
```
danielg0, line 817: I have an idea on how to maybe get this done in the db, can be done in another pr

danielg0 (commented):
Would be nice to have some tests for these new funcs, but understand that's difficult

internal/broadcast/types.go
YRMYJASKA: apparently, power draw is in Watts, so according to that we should be getting Joules. I will document that

internal/database/postgres.go
YRMYJASKA: we dont need to do that, because the output is agnostic of the order (confirm @NathanielB123 ?) and we stuff the GPUs by their timestamp either way

NathanielB123: > we dont need to do that, because the output is agnostic of the order (confirm @NathanielB123 ?)

I don't think this is true? I draw a line for each gpu, so if gpu order swaps, lines will use the wrong points

YRMYJASKA: currently the data is given out in `[(Timestamp, [GPU])]`. Does the ordering of `[GPU]` matter? if so, this needs to be fixed. If not, then things should work out alright

NathanielB123: Yes, the ordering does matter (again - line per GPU - if the `[GPU]` order swaps, then the points for each line swap at that point - if the happens multiple times, the graph is wrong)

YRMYJASKA: oh okay, then this needs to be fixed as per @danielg0's comment

PR #243, opened 2024-03-04
DRAFT: Adding files feature

Euphoride:
Solves #21, still draft, welcoming reviews keeping in mind TODOs need to be addressed.


frontend/tsconfig.json
```diff
@@ -14,7 +14,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react-jsx"
```
NathanielB123: Got thrown by this as well in https://github.com/gpuctl/gpuctl/pull/239. I *think* if you `npm update` then prettier will change it's mind and want the trailing commas gone.

YRMYJASKA (commented):
LGTM on the serverside stuff, cannot comment on the react code at all

frontend/src/Components/GroupInfoManagement.tsx
```diff
@@ -59,6 +76,110 @@ export const GroupInfoManagement = ({
...
+              const downloadLink = document.createElement("a");
+              downloadLink.href = URL.createObjectURL(
+                new Blob([await r.arrayBuffer()]),
+              );
+              downloadLink.download = currentFile;
+              downloadLink.click();
+            }
+          },
+          failure: () => {
+            // alert("Failure");
```
NathanielB123, line 99: Is there a chance we could create a `toast` here to inform the user that something went wrong? (see `useAddMachine` in `Hooks/Hooks.tsx` for an example)

EDIT: Oops, forgot the semantics of `useAuthFetch`. An invalid response is still considered a `success` so the toast should be above (in the case the `status != 200` but in any case - `toast`s on failed requests are probably a good idea.

EDIT 2: Actually, might be better actually to put the toast logic higher up in `Hooks.tsx` (consistent with `useAddMachine` etc... and means that if we use the hook in multiple places the toast logic doesn't need to be duplicated).

frontend/src/Components/GroupInfoManagement.tsx
```diff
@@ -59,6 +76,110 @@ export const GroupInfoManagement = ({
...
+    const downloader = useGetSpecificFile(
+      currentMachine,
+      currentFile,
+      (response) => {
+        validatedElim(response, {
+          success: async (r) => {
+            // here the response should have given an attached file through the attachment header stuff
+            // let's download
+
+            if (r.status == 200) {
```
NathanielB123, line 89: NIT: Elsewhere in the codebase we are using `===`

frontend/src/Components/GroupInfoManagement.tsx
```diff
@@ -59,6 +76,110 @@ export const GroupInfoManagement = ({
...
+    };
+  };
+
+  useEffect(() => {
+    if (currentMachine) {
+      view();
+    }
+  }, [currentMachine]);
+
+  // ...
```
NathanielB123, line 149: NIT: What does this comment mean?

frontend/src/Hooks/Hooks.tsx
```diff
@@ -2,14 +2,95 @@ import { useToast } from "@chakra-ui/react";
...
+  const len = file.byteLength;
+  for (let i = 0; i < len; i++) {
+    binaryString += String.fromCharCode(file[i]);
+  }
+  return btoa(binaryString);
+};
+
+export const useUploadFile = (callback: (r: Validated<Response>) => void) => {
+  const { useAuthFetch } = useAuth();
+  const [, uploadFileAuth] = useAuthFetch(UPLOAD_FILE_URL, callback);
```
NathanielB123, line 48: Toast?

frontend/src/Hooks/Hooks.tsx
```diff
@@ -2,14 +2,95 @@ import { useToast } from "@chakra-ui/react";
...
+        mime,
+        filename,
+        file_enc: encodeFile(file),
+      }),
+    });
+};
+
+export const useRemoveFile = () => {
+  const { useAuthFetch } = useAuth();
+  const [, removeFileAuth] = useAuthFetch(REMOVE_FILE_URL);
```
NathanielB123, line 67: Toast?

frontend/src/Hooks/Hooks.tsx
```diff
@@ -2,14 +2,95 @@ import { useToast } from "@chakra-ui/react";
...
+
+export const useGetSpecificFile = (
+  hostname: string,
+  filename: string,
+  callback: (r: Validated<Response>) => void,
+) => {
+  const { useAuthFetch } = useAuth();
+  const [, getSpecificFileAuth] = useAuthFetch(
+    GET_SPECIFIC_FILE_URL + "?hostname=" + hostname + "&file=" + filename,
+    callback,
```
NathanielB123, line 86: Toast?

NathanielB123 (approved):
Happy to merge now if you want. Just a couple nitpicks and ideally there would be some feedback to the user if the requests fail (`toast`s), but that can probably be addressed in a follow-up PR (i.e. if we want to try and merge this quickly to test in deployment).

PR #245, opened 2024-03-04
Change how install directory is derived

danielg0:
Rather than just using the data directory, combine it with the username we're ssh-ing in as too. This avoids permission issues when subsequent deployments change the user whose ssh key we're using and fail trying to write the new satellite binary to that folder.

### DIFF
--- a/internal/tunnel/onboard.go
+++ b/internal/tunnel/onboard.go
@@ -29,6 +29,13 @@ type Config struct {
 	KeyCallback ssh.HostKeyCallback
 }
 
+// install dir data dir appended with the username
+// this covers issues with subsequent deployments as different users not
+// having permissions to overwrite old data directories
+func (conf Config) installDir() string {
+	return conf.DataDir + conf.User
+}
+
 // Onboard will copy over and start the satellite on a remote machine, via SSH.
 //
 // - hostname must be an amd64 linux system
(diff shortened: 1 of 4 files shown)


internal/tunnel/onboard.go
```diff
@@ -29,6 +29,13 @@ type Config struct {
 	KeyCallback ssh.HostKeyCallback
 }
 
+// install dir data dir appended with the username
+// this covers issues with subsequent deployments as different users not
+// having permissions to overwrite old data directories
+func (conf Config) installDir() string {
+	return conf.DataDir + conf.User
```
sirjacobofglais: could we do dir_user? feels like this might be confusing in some cases

aDotInTheVoid: NIT: This should use https://pkg.go.dev/path#Join

internal/tunnel/onboard.go
```diff
@@ -47,8 +54,8 @@ func Onboard(
 	}
 	defer client.Close()
 
-	// -- Make the data dir --
-	err = runCommand(client, "mkdir -p "+conf.DataDir)
+	// -- Make the install dir --
+	err = runCommand(client, "mkdir -p "+conf.installDir())
```
aDotInTheVoid, line 60: Could we rename the `DataDir` field to `DataDirTemplate`, so it's clear that it shouldn't be used on it's own?

PR #247, opened 2024-03-05
Increase RAM during NPM build

danielg0:
We were getting out-of-memory crashes whilst building the frontend, so increase the memory limit.


NathanielB123 (approved):
LGTM!

PR #249, opened 2024-03-05
all: Correct use of time

aDotInTheVoid:
Means the config is readable, and should fix dead-machine-monitor.

r? @danielg0 

cc @Euphoride

### DIFF
--- a/internal/config/duration.go
+++ b/internal/config/duration.go
@@ -1,0 +1,35 @@
+package config
+
+import "time"
+
+type Duration time.Duration
+
+const (
+	Nanosecond  Duration = Duration(time.Nanosecond)
+	Microsecond Duration = Duration(time.Microsecond)
+	Millisecond Duration = Duration(time.Millisecond)
+	Second      Duration = Duration(time.Second)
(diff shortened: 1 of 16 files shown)


internal/config/duration.go
```diff
@@ -0,0 +1,35 @@
+package config
+
+import "time"
+
+type Duration time.Duration
+
+const (
```
Euphoride, line 7: NIT/Curiosity: Why redefine?

Euphoride (commented):
LGTM dependent on CI passing - some churn on variable naming but readable PR anyhow

internal/config/duration.go
aDotInTheVoid: So that their of the right type.

danielg0 (approved):
LGTM, @aDotInTheVoid, tests pass now. Do you mind if I make the couple of other changes needed to remove the other references to Unixtimes in postgres.go, or do you want to leave this to another pr?

aDotInTheVoid:
Let’s do that in a follow up, this is already kinda big

PR #250, opened 2024-03-05
Add admin panel editable fields to big card view

sirjacobofglais:
Only shown when logged in as admin

### DIFF
--- a/frontend/src/Components/EditableFields.tsx
+++ b/frontend/src/Components/EditableFields.tsx
@@ -22,10 +22,8 @@ export const EditableField = ({
   workstation,
   fieldKey,
   placeholder,
-  GroupSelect,
   isEven,
 }: EditableFieldProps & {
-  GroupSelect: GS;
   isEven: boolean;
 }) => {
   const pickCol = (value: string) => (value ? "gray.600" : "gray.300");
--- a/frontend/src/Components/WorkstationView.tsx
+++ b/frontend/src/Components/WorkstationView.tsx
@@ -26,7 +26,7 @@ import { useSearchParams } from "react-router-dom";
 import { Graph } from "./Graph";
 import { useHistoryStats } from "../Hooks/Hooks";
 import { useState } from "react";
-import { GraphField } from "../Data";
+import { GraphField, WorkStationData, WorkStationGroup } from "../Data";
 import {
   Validation,
   all,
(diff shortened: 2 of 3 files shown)


frontend/src/Components/EditableFields.tsx
```diff
@@ -22,10 +22,8 @@ export const EditableField = ({
   workstation,
   fieldKey,
   placeholder,
-  GroupSelect,
   isEven,
 }: EditableFieldProps & {
-  GroupSelect: GS;
```
NathanielB123, line 28: Yeah removing this is probably fair. I added it to support group selection being a drop-down rather than pure text input, and I know at one point I had the intention to go back to drop-downs but I think it's probably not worth doing now.

frontend/src/Components/WorkstationView.tsx
```diff
@@ -57,42 +59,60 @@ const GRAPH_FIELDS = enumVals(GraphField);
...
-            <StatsTable hostname={hostname}></StatsTable>
-            <StatsGraph hostname={hostname} />
-          </SimpleGrid>
-        </ModalBody>
-      </ModalContent>
-    </Modal>
-  );
+  return validationElim(allStats, {
+    success: (stats) => {
+      const { wstat, name } = stats.flatMap((g) =>
```
NathanielB123: If someone deletes the workstation we are on the page for, I think the `[0]` will error. Do we want to handle this case (ig deselect the workstation)?

frontend/src/Components/WorkstationView.tsx
```diff
@@ -123,51 +143,30 @@ const StatsTable = ({ hostname }: { hostname: string }) => {
...
-        },
-        failure: () => <Text>Failure fetching data! Retrying...</Text>,
-        loading: () => <Text>Fetching data...</Text>,
-      })}
+      <TableContainer>
+        <Table variant="striped">
+          <Thead>
+            <Tr>
+              <Th>Field</Th>
+              <Th>GPU 0</Th>
```
NathanielB123, line 158: Oh shit. Not directly relevant to this PR but I forgot I never fixed the headings here (should be as many as there are GPUs connected to that workstation - I think the actual rows should be fine?)

frontend/src/Components/WorkstationView.tsx
NathanielB123: Opened an issue https://github.com/gpuctl/gpuctl/issues/251

NathanielB123 (approved):

frontend/src/Components/WorkstationView.tsx
NathanielB123: Opened an issue https://github.com/gpuctl/gpuctl/issues/251

frontend/src/Components/EditableFields.tsx
sirjacobofglais: Yeah the overhead here was causing problems for reusing this in the big card

PR #255, opened 2024-03-05
Make UUIDs not strings

### LINKED ISSUE #129: UUID not a string pls

danielg0:
Move to using Google's uuid library and  uuid in the postgres database.

The fake gpu generates two random uuids on each use. This should also
mean we don't run into issues when two different hosts are both running
under fake gpu data I've removed the fake gpu tests, because it was only
checking the uuid matched.

Closes https://github.com/gpuctl/gpuctl/issues/129


internal/gpustats/fake_test.go
```diff
@@ -1,17 +0,0 @@
-package gpustats
-
-import (
-	"testing"
-
-	"github.com/stretchr/testify/assert"
-)
-
-// IMMENSELY STUPID WE HAVE TO WRITE TESTS FOR THESE
```
aDotInTheVoid, line 9: LGF!!!

aDotInTheVoid (approved):
LGTM, gonna let you merge this as it looks like it requires migrating or nuking the prod DB

PR #260, opened 2024-03-06
fixed satellite config file generation

YRMYJASKA:
- Made `config.Duration` into an alias for `time.Duration` as otherwise toml cannot figure out that we are indeed using time here
- fixed broken test for config file marshalling


internal/config/duration.go
```diff
@@ -13,7 +13,7 @@ const (
 	Hour        Duration = Duration(time.Hour)
 )
 
-func (d *Duration) UnmarshalText(text []byte) error {
```
YRMYJASKA, line 16: These functions seem to not be used at all. Can we delete `UnmarshalText` and `NewTicker`? @aDotInTheVoid

aDotInTheVoid:
My understanding is that toml wouldn't unmarshal from text into a `time.Duration`, so we needed our own type that would allow it. Although from this it looks like we can still parse the human readable stuff. Given that, I'd rather just delete the type alias and use `time.Duration` everywhere.

Test changes look good.

YRMYJASKA:
> My understanding is that toml wouldn't unmarshal from text into a `time.Duration`, so we needed our own type that would allow it. Although from this it looks like we can still parse the human readable stuff. Given that, I'd rather just delete the type alias and use `time.Duration` everywhere.
> 
> Test changes look good.

~~According to here from the docs, we can parse into `time.Duration` automatically: https://pkg.go.dev/github.com/burntsushi/toml#readme-using-the-encoding-textunmarshaler-interface~~ I need to learn reading comprehension, smh. I will mull over this

YRMYJASKA:
Interesting, the docs are out of date it seem. https://github.com/BurntSushi/toml/blob/0e879cbdab1075e5622c5c91581eb5e417e6fc1a/decode_test.go#L808 The library nowadays supports `time.Duration` and the example I linked above is now replaced with an email example in the README. Good for us, we can ditch `config.Duration` completely.

PR #262, opened 2024-03-06
remove installation dir on deboard

### LINKED ISSUE #244: Removing the data directory on deboard
In an attempt to stay on CSG's good side, it's probably worth removing the existing data directory when we deboard a machine

YRMYJASKA:
Seems like a small change. Not sure if anything else is needed here.
Closes #244


danielg0 (approved):
LGTM

PR #292, opened 2024-03-07
deploy: Run control as multi-stage alpine build.

aDotInTheVoid:
This makes images smaller, and means that we have the ping binary on the running
machine.


danielg0 (approved):
LGTM!

PR #294, opened 2024-03-07
Refactor database query to handle conflicts in Stats table

Euphoride:
Title.


danielg0 (approved):
Downsample could use tests now we have historical stats, but that can be another pr. LGTM

PR #297, opened 2024-03-07
EMERGENCY DOWNSAMPLER

YRMYJASKA:
in case we dont get downsampling working well, every stat sample will expire after a week (threshold downsampling)


NathanielB123:
Converted to a draft to prevent accidentally merging without thinking (like I *almost* just did)

Euphoride:
Closed due to downsampler fixes

PR #298, opened 2024-03-07
Tidy up group & info management

### LINKED ISSUE #278: css on admin table
when you open the editable, the side jumps to the size of the column header. make good.

danielg0:
Give items a minimum width so editing doesn't shrink it (closes #278). Also change the spacing and padding to make it a bit tighter and vaguelly usable on phones.

Make the note popover bigger

Group actions into one column


NathanielB123 (approved):
LGTM!

PR #304, opened 2024-03-07
Make odd/even rows appear in alternating colours

danielg0:
We have a row in a link, rather than a link in a row. Also take the opportunity to extract out the row and add some interactivty on hover, to display the fact a link exists


NathanielB123 (approved):
LGTM!

PR #306, opened 2024-03-07
Pass $1 param and set downsample thresh to 1 hour

NathanielB123:


NathanielB123:
(Daniel has verbally approved)

PR #312, opened 2024-03-08
Update SSH command to reboot instead of shutdown

Euphoride:
Title.


danielg0 (changes requested):
`shutdown` takes a time, but reboot doesn't, so this command should be just `reboot`

Euphoride:
> `shutdown` takes a time, but reboot doesn't, so this command should be just `reboot`

This has been addressed in the latest commit

danielg0 (commented):
LGTM

danielg0 (approved):

PR #313, opened 2024-03-08
Added downsample/delete config

Euphoride:
Title.

### DIFF
--- a/internal/config/groundstation.go
+++ b/internal/config/groundstation.go
@@ -24,8 +24,8 @@ type Database struct {
 	Postgres            bool          `toml:"postgres"`
 	PostgresUrl         string        `toml:"url"`
 	DownsampleInterval  time.Duration `toml:"downsample_interval"`
-	DownsampleThreshold time.Duration `toml:"downsample_threshold"`
-	DeleteThreshold     time.Duration `toml:"delete_threshold"`
+	DownsampleThreshold time.Duration `toml:"downsample_threshold"` // Downsample threshold is the time after which data is downsampled
+	DeleteThreshold     time.Duration `toml:"delete_threshold"`     // Delete threshold is the time after which data is deleted from the database
 }
 
 type AuthConfig struct {
--- a/internal/config/groundstation.go
+++ b/internal/config/groundstation.go
@@ -20,11 +20,12 @@ func (t Timeouts) MonitorInterval() time.Duration {
 }
 
 type Database struct {
-	InMemory           bool          `toml:"inmemory"`
-	Postgres           bool          `toml:"postgres"`
-	PostgresUrl        string        `toml:"url"`
-	DownsampleInterval time.Duration `toml:"downsample_interval"`
-	DownsampleType     string        `toml:"downsample_type"`
+	InMemory            bool          `toml:"inmemory"`
+	Postgres            bool          `toml:"postgres"`
+	PostgresUrl         string        `toml:"url"`
+	DownsampleInterval  time.Duration `toml:"downsample_interval"`
+	DownsampleThreshold time.Duration `toml:"downsample_threshold"`
+	DeleteThreshold     time.Duration `toml:"delete_threshold"`
 }
 
 type AuthConfig struct {
@@ -54,10 +55,11 @@ func DefaultControlConfiguration() ControlConfiguration {
 			WAPort: 8000,
 		},
 		Database: Database{
-			InMemory:       false,
-			Postgres:       false,
-			PostgresUrl:    "postgres://postgres@postgres/postgres",
-			DownsampleType: "DOWNSAMPLE",
+			InMemory:            false,
+			Postgres:            false,
+			PostgresUrl:         "postgres://postgres@postgres/postgres",
+			DownsampleThreshold: 1 * time.Hour,
+			DeleteThreshold:     24 * time.Hour,
 		},
 		Auth: AuthConfig{
 			Username: "admin",
--- a/internal/config/groundstation.go
+++ b/internal/config/groundstation.go
@@ -24,6 +24,7 @@ type Database struct {
 	Postgres           bool          `toml:"postgres"`
 	PostgresUrl        string        `toml:"url"`
 	DownsampleInterval time.Duration `toml:"downsample_interval"`
+	DownsampleType     string        `toml:"downsample_type"`
 }
 
 type AuthConfig struct {
@@ -53,9 +54,10 @@ func DefaultControlConfiguration() ControlConfiguration {
 			WAPort: 8000,
 		},
 		Database: Database{
-			InMemory:    false,
-			Postgres:    false,
-			PostgresUrl: "postgres://postgres@postgres/postgres",
+			InMemory:       false,
+			Postgres:       false,
+			PostgresUrl:    "postgres://postgres@postgres/postgres",
+			DownsampleType: "DOWNSAMPLE",
 		},
 		Auth: AuthConfig{
 			Username: "admin",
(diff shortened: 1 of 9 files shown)


internal/config/groundstation.go
```diff
@@ -53,9 +54,10 @@ func DefaultControlConfiguration() ControlConfiguration {
 			WAPort: 8000,
 		},
 		Database: Database{
-			InMemory:    false,
-			Postgres:    false,
-			PostgresUrl: "postgres://postgres@postgres/postgres",
+			InMemory:       false,
+			Postgres:       false,
+			PostgresUrl:    "postgres://postgres@postgres/postgres",
+			DownsampleType: "DOWNSAMPLE",
```
danielg0: What's the aim with adding a downsample types? You've defined two, "DELETE" & "DOWNSAMPLE" (which could be more descriptive), but I think the end goal is for us to use a combination of both (eg. downsample result from more than a day ago, then delete all samples from more than a month ago)

internal/database/prune.go
```diff
@@ -5,14 +5,24 @@ import (
 	"time"
 )
 
-func DownsampleOverTime(interval time.Duration, database Database) error {
+func DownsampleOverTime(interval time.Duration, downsample_type string, database Database) error {
 	downsampleTicker := time.NewTicker(time.Duration(interval))
 
 	for range downsampleTicker.C {
-		err := database.Downsample(time.Now())
+		if downsample_type == "DOWNSAMPLE" {
```
danielg0: I think the different Downsample types should be constant strings in the database interface file, rather than magic strings

danielg0 (commented):
Some comments on the code. Overall, I'm not sure which problem this solves and I'm not sure if this is the simplest approach - we're likely going to end up using a combination of combining old samples and deleting really old samples, so having delete accessible through the downsample function and the public Delete() function seems like a recipe for accidentally calling delete or downsample more than/not when we intend to.

Euphoride:
> Some comments on the code. Overall, I'm not sure which problem this solves and I'm not sure if this is the simplest approach - we're likely going to end up using a combination of combining old samples and deleting really old samples, so having delete accessible through the downsample function and the public Delete() function seems like a recipe for accidentally calling delete or downsample more than/not when we intend to.


We'll have to loop back with what Ben specifically wanted. He leans towards outright deletion, and either way, if we want a hybrid approach, we will need both methods.

internal/config/groundstation.go
Euphoride: These are as descriptive as they get, that said, the configuration has changed to reflect more fine-tunable configurations.

internal/config/groundstation.go
```diff
@@ -20,10 +20,12 @@ func (t Timeouts) MonitorInterval() time.Duration {
...
 type Database struct {
-	InMemory           bool          `toml:"inmemory"`
-	Postgres           bool          `toml:"postgres"`
-	PostgresUrl        string        `toml:"url"`
-	DownsampleInterval time.Duration `toml:"downsample_interval"`
+	InMemory            bool          `toml:"inmemory"`
+	Postgres            bool          `toml:"postgres"`
+	PostgresUrl         string        `toml:"url"`
+	DownsampleInterval  time.Duration `toml:"downsample_interval"`
+	DownsampleThreshold time.Duration `toml:"downsample_threshold"`
```
aDotInTheVoid: These should probably have some form of docs attached to them.

internal/database/interface.go
```diff
@@ -45,7 +45,8 @@ type Database interface {
 	UpdateMachine(changes broadcast.ModifyMachine) error
 
 	// downsample since certain time
-	Downsample(time.Time) error
+	Downsample(time.Duration) error
+	Delete(time.Duration) error
```
aDotInTheVoid: This needs a name that says what it does. Something like `DeleteOldStats`.

internal/database/interface.go
```diff
@@ -45,7 +45,8 @@ type Database interface {
 	UpdateMachine(changes broadcast.ModifyMachine) error
 
 	// downsample since certain time
```
aDotInTheVoid: This comments needs to be updated.

internal/database/postgres.go
```diff
@@ -221,7 +221,9 @@ func (conn PostgresConn) UpdateGPUContext(host string, packet uplink.GPUInfo) er
 	return err
 }
 
-func (conn PostgresConn) Downsample(int_now time.Time) error {
+func (conn PostgresConn) Downsample(thresh time.Duration) error {
+	int_now := time.Now()
```
aDotInTheVoid: NIT: This should be `intNow`. (And similarly, all the rest of the code should follow go's style)

internal/database/inmemory.go
```diff
@@ -161,21 +161,21 @@ func (m *inMemory) LastSeen() ([]broadcast.WorkstationSeen, error) {
...
-			if sample.Time < cutoffTime.Unix() {
+			if sample.Time < time.Now().Add(cutoffTime).Unix() {
 				oldSamples = append(oldSamples, sample)
 			} else {
 				newSamples = append(newSamples, sample)
 			}
 		}
 
-		sort.Slice(oldSamples, func(i, j int) bool {
+		sort.Slice(oldSamples, func(i int, j int) bool {
```
aDotInTheVoid, line 178: NIT: Could we not change this for no reason?

internal/database/prune.go
```diff
@@ -1,24 +1,20 @@
 package database
 
 import (
-	"log/slog"
 	"time"
 )
 
-func DownsampleOverTime(interval time.Duration, database Database) error {
+func DownsampleOverTime(interval time.Duration, downsample_thresh time.Duration, delete_thresh time.Duration, database Database) error {
```
aDotInTheVoid: It's probably nicer to make a new struct here, so it's clearer at the callsight the args are right.

```go
type PruneConfig struct {
   interval time.Duration
   deleteThreshold time.Duration
   downsampeThreshold time.Duration
}
```

internal/database/prune.go
```diff
@@ -1,24 +1,20 @@
...
 	downsampleTicker := time.NewTicker(time.Duration(interval))
 
 	for range downsampleTicker.C {
-		err := database.Downsample(time.Now())
-
-		if err != nil {
```
aDotInTheVoid, line 14: Why was this removed?

internal/database/prune_test.go
```diff
@@ -136,21 +136,19 @@ func TestAverageProcess(t *testing.T) {
 
 // specify the time we test at
 // test hasn't been passing since March started?
-var now = time.Date(2024, time.February, 1, 0, 0, 0, 0, time.Local)
 
 func TestDownsample(t *testing.T) {
+	var now = time.Now()
```
aDotInTheVoid: NIT: This should be `now := time.Now()`, no?

internal/database/prune.go
```diff
@@ -1,24 +1,20 @@
...
-			slog.Error("Got error whilst downsampling", "err", err)
-		}
+		database.Downsample(-1 * downsample_thresh)
```
aDotInTheVoid: NIT: Why not just `-downsample_thresh` here. (And loads of other places in this PR)

aDotInTheVoid (commented):
I'm not sure if this is the right approach to downsampling, as i'm not framiliar enough with the domain, but I've given some feedback that's hopefully usefull regardless.'

I'd like another set of eyes on this before it goes it, as Downsampling has been a menace, and I don't want to break it so close to the deadline.

PR #314, opened 2024-03-08
Add sorting by column functionality

Euphoride:
Title.


frontend/src/Providers/FetchProvider.tsx
```diff
@@ -72,9 +72,10 @@ export const FetchStatsProvider = ({
 const retrieveAllStats: (
   cb: StatsCB,
 ) => Promise<Validated<WorkStationGroup[]>> = async (cb) => {
-  const stats = success(
-    preProcess(await (await fetch(API_URL + API_ALL_STATS_PATH)).json()),
-  );
+  const stats = success(preProcess(EXAMPLE_DATA_1));
```
NathanielB123: For merging to trunk can we change this back pls?

NathanielB123 (changes requested):
We should change frontend back to use the webapi to collect data - otherwise I'm happy

NathanielB123 (approved):

PR #319, opened 2024-03-09
Fix changing the height of the notes box

danielg0:


danielg0 (commented):
Approved verbally by @sirjacobofglais

PR #328, opened 2024-03-10
Optimise aggregate query

danielg0:
Rather than calculating voltage in Go code, do everything in the postgres query using aggregate operators.

### DIFF
--- a/cmd/control/control.go
+++ b/cmd/control/control.go
@@ -80,9 +80,9 @@ func main() {
 	}
 	totalEnergy.Store(data.TotalEnergy)
 
-	const cache_duration = time.Hour
+	const cacheDuration = time.Hour
 	go func() {
-		for range time.NewTicker(cache_duration).C {
+		for range time.NewTicker(cacheDuration).C {
 			data, err := db.AggregateData()
 			if err != nil {
 				log.Error("Got error calculating new cache value of aggregate", "err", err)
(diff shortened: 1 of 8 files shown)


cmd/control/control.go
```diff
@@ -69,8 +71,28 @@ func main() {
...
+	totalEnergy.Store(data.TotalEnergy)
+
+	const cache_duration = 15 * time.Minute
+	go func() {
+		for range time.NewTicker(cache_duration).C {
+			data, err := db.AggregateData()
+			if err != nil {
+				log.Error("Got error calculating new cache value of aggregate", "err", err)
+			}
+			totalEnergy.Store(data.TotalEnergy)
```
aDotInTheVoid: This should be in the `else` branch.

cmd/control/control.go
```diff
@@ -69,8 +71,28 @@ func main() {
...
+	// calculating aggregate data takes a while, so cache it
+	var totalEnergy atomic.Uint64
+	data, err := db.AggregateData()
+	if err != nil {
+		log.Error("Got error calculating initial value of aggregate", "err", err)
+		os.Exit(-1)
+	}
+	totalEnergy.Store(data.TotalEnergy)
+
+	const cache_duration = 15 * time.Minute
```
aDotInTheVoid: NIT: This should use `cammelCase`

PR #332, opened 2024-03-11
cooking with sql

YRMYJASKA:


danielg0:
Why do we have to get rid of all the query formatting at the same time?

PR #333, opened 2024-03-11
In postgres, rounding doesn't cast

danielg0:


NathanielB123:
Solved by https://github.com/gpuctl/gpuctl/pull/332

